PR #24, opened 2021-04-26
[Nacho-0050] Option to disable enchantment table ticking

HeathLoganCampbell:
Thought this would be useful for servers that are basically just hubs or don't need the functionality


Lucaskyy:
LGTM, merging.

PR #41, opened 2021-05-06
KigPaper-0172 Dont decode block place item just use item in hand

HeathLoganCampbell:


Lucaskyy:
Heath going ham with the PRs haha, LGTM!

I will merge the commits from GreenMC first to make sure there are no merge conflicts :)

Lucaskyy:
LewUwU recommended me to not merge this patch because there's a chance it could break a lot of plugins that utilize NMS and this packet. I could add this patch with a config option so if players have issues with this they could disable it.

I think it might be a good idea then to disable it by default but let players enable it at their own risk.

LewUwU:
Merge and do a config option disabled by default to enable this patch

CyberFlameGO (changes requested):
Add configuration functionality as this is a potentially breaking change

andreasdc:
> LewUwU recommended me to not merge this patch because there's a chance it could break a lot of plugins that utilize NMS and this packet. I could add this patch with a config option so if players have issues with this they could disable it.
> 
> I think it might be a good idea then to disable it by default but let players enable it at their own risk.

What is the risk?

PR #61, opened 2021-06-09
Java 16

Sneakometer:
There is a warning printed to log when running the server using java 16 even though it's fully compatible.
I'm running this for some hours in production now and haven't seen any issues.

### DIFF
--- a/NachoSpigot-Server/pom.xml
+++ b/NachoSpigot-Server/pom.xml
@@ -49,7 +49,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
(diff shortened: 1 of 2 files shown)


LewUwU:
Can you send the warning?

Sneakometer:
It's directly below the chaned if statement from the commit.

`System.err.println("It seems like you are not using Java 15!");`

CyberFlameGO (approved):
LGTM - @Lucaskyy @HeathLoganCampbell merge?

LewUwU:
I would like fix the print issue before merging

Lucaskyy:
Which print issue?

PR #62, opened 2021-06-09
Revert commit 08eb620bd73fc1aa537ccacfae1ed19bbddbf7da

bob7l:
There's no purpose for using a Set here. The "players" collection shouldn't contain duplicates, and no where in the code do we use "disabledFlushes" in a context that would benefit from HashSet. The ArrayList is both faster in this scenario, and makes more sense.

### DIFF
--- a/NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
+++ b/NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
@@ -876,7 +876,7 @@ public abstract class MinecraftServer implements Runnable, ICommandListener, IAs
             if(this.getPlayerList().getPlayerCount() != 0) // Tuinity
             {
                 // Tuinity start - controlled flush for entity tracker packets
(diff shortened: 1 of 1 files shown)


Sneakometer:
But it says "Heavily optimize"

Lucaskyy:
@LewUwU This is your commit, want to merge this or not?

HeathLoganCampbell:
@Lucaskyy Vouch for anything Bob7I related, guy is basically built like a god

LewUwU:
Mojang allows duplicated names, and HashSet is a little more speedy than ArrayList in some cases, i tested and see a 12.3% of performance improvement with 230 players with the use of HashSet

Lucaskyy:
Making a single network manager for all users would probably be so much better because less objects in memory but man oh man will that break every plugin out there haha.

Lucaskyy:
@LewUwU @bob7l So are we merging?

CyberFlameGO (approved):
LGTM

HeathLoganCampbell (approved):

PR #63, opened 2021-06-10
Backported 1.12.2 Keep Alive handling

Outfluencer:
fixed the issue that players whitch not sending keep alives don't get kicked
And the fact that not sending keep alives disables some checks of some anticheats


Lucaskyy:
You broke the KEEPALIVE_LIMIT variable, can you create a new PR to fix it?

Outfluencer:
i did

PR #66, opened 2021-06-10
Update ViaVersion Packages

RealBauHD:
The packages have been changed in the latest ViaVersion version.


Outfluencer:
thats right needs to be merged

Lucaskyy:
This won't work though for people using older versions of ViaVersion. Try checking first if the class exists, and then decide which class prefix to use.

CyberFlameGO (approved):
LGTM - adding a check for older versions would be beneficial though

Lucaskyy:
Oh? Why close it?

crafter23456:
> Oh? Why close it?

It got implemented in another PR #70

Issue #64, opened 2021-06-10
Could not call method 'public static org.bukkit.inventory.meta.ItemMeta


AltarRiza:
Hello i use strikepractice with default setting. https://pastebin.ubuntu.com/p/zCC8G3TC5j/ . With paper or spigot does not give this error

Lucaskyy:
Looks like an easy fix to me, I will do that when I'm home.

AltarRiza:
I am waiting

AltarRiza:
I'm still waiting

CyberFlameGO:
Please be a little more patient, but here. @Lucaskyy Boop!

PR #70, opened 2021-06-12
Implimented the ViaVersion patch as wanted & Fixed NPE

Outfluencer:
Checking for new and old versions of viaversion and fixed NPE in SkullMeta 

The skinCache impl from Paper causes  the #64 issue


Lucaskyy:
Thank you for this PR! Let me know when I'm able to merge it.

Outfluencer:
You can merge it, don't want to change more

PR #71, opened 2021-06-12
Java 15 -> Java 16

crafter23456:
I think he forgot it in commit https://github.com/CobbleSword/NachoSpigot/commit/e8e3ee2cd4041630351951eede75335f621baf51


Lucaskyy:
No. I am not merging this because Java 16 isn't natively supported and some things are broken in it like ProtocolLib. Java 15 is still the way to go when you can.

PR #73, opened 2021-06-13
Found a issue in 1.12.2 keep alive handling

Outfluencer:
A client can send the same keep alive packet 100 times and does not get  kicked.
Checking if isPendingPing is true is fixing this issue, mc 1.12.2 does not fixed this, but i found this myself.
And you can influence your ping by spamming these packets


Lucaskyy:
Thank you for this PR! :)

PR #82, opened 2021-06-21
Various non-breaking changes

CyberFlameGO:
Formatting + console messages


CyberFlameGO:
Resolves https://github.com/CobbleSword/NachoSpigot/pull/61#issuecomment-858511283

Lucaskyy:
I won't merge this because Java 16 is actually not supported. It breaks Reflection which breaks all patches. The rest aren't really of use, since they don't do anything. Is there a reason behind this or do you agree?

CyberFlameGO:
Ah fair enough, I agree

PR #90, opened 2021-06-29
Fix explosions when called via the api

ad1tya2:


CyberFlameGO (approved):
I don't understand the point of this but the changes aren't breaking by the looks of it, so LGTM.

ad1tya2:
so the thing is, When using the craftworld api to create an explosion, due to [this line of code specifically](https://github.com/CobbleSword/NachoSpigot/blob/d0614b7e3fc1dab2c662f27f14188593daa17424/NachoSpigot-Server/src/main/java/org/bukkit/craftbukkit/CraftWorld.java#L569) There is a null pointer exception [here](https://github.com/CobbleSword/NachoSpigot/blob/5efae34642aff85fe72b4bb46491b550fecff16c/NachoSpigot-Server/src/main/java/net/minecraft/server/Explosion.java#L72) because source is null, so to avoid that i have added this check.
The Event is still triggered though due to [this](https://github.com/CobbleSword/NachoSpigot/blob/84c9607af603b1d99dff0903b21ee733d6afa652/NachoSpigot-Server/src/main/java/net/minecraft/server/World.java#L2130)

It took me like 3 hours to track this issue down completely and fix it xD,
TLDR: This shouldnt affect anything at all in theory and should fix the nullpointer issue when the createExplosion api is used

CyberFlameGO:
Ah I see. @Lucaskyy @HeathLoganCampbell can one of you merge this?

hpfxd (approved):

PR #99, opened 2021-07-19
FlamePaper-0117 Pearl through blocks

Rastrian:
Original Patch: https://github.com/2lstudios-mc/FlamePaper/blob/ver/1.8.8/Spigot-Server-Patches/0117-Pearl-through-blocks.patch

This patch add options to accept enderpearl bypass fence gates and tripwires.
The only change of this backport to the original is the complete fence gate list, coweb and bed block(s), which is not present on the flamepaper patch.


Lucaskyy:
Thank you for both PRs! :) <3

PR #102, opened 2021-07-19
rework networking

Lucaskyy:
reworked some of the networking, haven't tested it yet though.
some of the patches are from Minestom, possibly adding more.

added support for KQueue
made the packet buffer larger (going to add a config option for that)

and some other things :)

### DIFF
--- a/NachoSpigot-Server/src/main/java/net/minecraft/server/DedicatedServer.java
+++ b/NachoSpigot-Server/src/main/java/net/minecraft/server/DedicatedServer.java
@@ -2,8 +2,6 @@ package net.minecraft.server;
 
 import java.io.File;
 import java.io.IOException;
-import java.lang.reflect.Field;
-import java.lang.reflect.Method;
 import java.net.InetAddress;
 import java.net.Proxy;
 import java.util.Random;
@@ -13,7 +11,6 @@ import java.util.concurrent.TimeUnit;
 import dev.cobblesword.nachospigot.Nacho;
 import dev.cobblesword.nachospigot.commons.IPUtils;
 import dev.cobblesword.nachospigot.knockback.Knockback;
-import dev.cobblesword.nachospigot.patches.RuntimePatches;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -21,7 +18,6 @@ import org.apache.logging.log4j.Logger;
 import java.io.PrintStream;
 import org.apache.logging.log4j.Level;
 
-import org.bukkit.Bukkit;
 import org.bukkit.craftbukkit.LoggerOutputStream;
 import co.aikar.timings.SpigotTimings; // Spigot
 import org.bukkit.event.server.ServerCommandEvent;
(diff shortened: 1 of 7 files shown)


Lucaskyy:
soon a new repo will be created, which will utilize Golang and IPC to completely transfer incoming and outgoing packets to a much faster library in Go. Should allow for higher player counts :) this will break every plugin that utilizes Netty, but there are plans to "simulate" the Netty API in Java which then calls our Golang library using IPC.

Still have to do some research on IPC, on how fast it is, and if it's going to cause bottlenecks or not.

Opinions on that? @HeathLoganCampbell

NachoSpigot-Server/src/main/java/net/minecraft/server/DedicatedServer.java
```diff
@@ -459,7 +455,15 @@ public boolean ae() {
...
-        return this.propertyManager.getBoolean("use-native-transport", true) && org.apache.commons.lang.SystemUtils.IS_OS_LINUX;
+        return this.getTransport() == ServerConnection.EventGroupType.EPOLL && org.apache.commons.lang.SystemUtils.IS_OS_LINUX;
+    }
+
+    public ServerConnection.EventGroupType getTransport() {
+        try {
+            return ServerConnection.EventGroupType.valueOf(this.propertyManager.getString("transport-to-use", "default").toUpperCase());
+        } catch (Exception ignored) {
+            return ServerConnection.EventGroupType.DEFAULT;
+        }
```
hpfxd, line 466: I feel like the `use-native-transport` option should stay as it was, and use KQueue or Epoll depending on what platform the server is running on.

NachoSpigot-Server/src/main/java/net/minecraft/server/ServerConnection.java
```diff
@@ -1,50 +1,51 @@
...
 
-    private static boolean isUsingEpoll = false;
+    private static final WriteBufferWaterMark SERVER_WRITE_MARK = new WriteBufferWaterMark(1 << 20, 1 << 21);
 
-    public static LazyInitVar<NioEventLoopGroup> a;
-    public static LazyInitVar<EpollEventLoopGroup> b;
-    public static final LazyInitVar<DefaultEventLoopGroup> c = new LazyInitVar() {
-        protected DefaultEventLoopGroup a() {
-            return new DefaultEventLoopGroup(0, (new ThreadFactoryBuilder()).setNameFormat("Netty Local Server IO #%d").setDaemon(true).build());
-        }
```
hpfxd, line 38: This part should probably stay as it is. Last time I tried to remove the `LazyInitVar`s, it broke plugins such as ProtocolLib and ViaVersion.

hpfxd (changes requested):

Lucaskyy:
oops, I already merged this haha. that was not meant to happen

NachoSpigot-Server/src/main/java/net/minecraft/server/DedicatedServer.java
Lucaskyy: Normally its set to default, this then chooses if it should use Epoll, KQueue, or NIO if none of the others are supported. Why should we add that option back, if it's still there but now you need to do:
`transport-to-use: epoll` for epoll
`transport-to-use: kqueue` for kqueue
`transport-to-use: default` for epoll, kqueue, or NIO depending on your system

NachoSpigot-Server/src/main/java/net/minecraft/server/ServerConnection.java
Lucaskyy: I'll add them back.

NachoSpigot-Server/src/main/java/net/minecraft/server/DedicatedServer.java
hpfxd: This is from the default Minecraft `server.properties`, correct? I don't feel like we should be modifying the structure upstream configs
Also, I've never seen situations where both KQueue and Epoll are supported at the same time, so I'm not sure how useful the option is

Also, should probably add NIO as an `EventGroupType` if it does stay, as it doesn't look like there's a way to explicitly select it

Lucaskyy: I'll add something for that 👍 Please see the new PR where I continue #104

PR #103, opened 2021-07-19
New Configs for optional performance gain

Rastrian:
Added new configs options for optional performance gain, this is an easy and performatic way to allow custom servers change their configs without plugin throttling, and also, give an option to the servers administrators ajust the configurations under their desire.

About the configurations implemented:
- infiniteWaterSources: control the behavior of infinite water method.
- leavesDecayEvent: control the behavior of leavesDecayEvent event, even before it fires inside the code.
- enableMobAI: controls MobAI
- enableMobSound: controls mobs sounds.
- enableEntityActivation: control the behavior of entity activation.
- enableLavaToCobblestone: control the behavior into transforming lava to cobblestone blocks, and also obsidian.
- firePlayerMoveEvent: control the event PlayerMove, if this config be disabled, the server will not track the player movement, can be helpful for performance, but also, can be dangerous for some plugins implementations.
- endermiteSpawning: control the spawn of endermite after throw an enderpearl.
- disablePhysicsPlace: control the physics behavior of a block after place it (example: sand).
- disablePhysicsUpdate: control the physics behavior of a block after it change/update (example: when you break an sand block in the desert and the blocks fall, if this option be disabled, this will not happen).
- doBlocksOperations: control the operation of blocks on the server (example: the growth of a cane or wheat crop may be stopped if this setting is turned off).
- doChunkUnload: control the behavior of the chunk unload, this can be helpful for lobby/limbo servers.
- chunkThreads: control the amount of threads designed to manage the chunks, if you change this option we recommend you change the netty-threads in spigot.yml (default by paper: 2).
- playersPerThread: control the amount of players designed to be in a thread (default by paper: 50).

On the future i will be implementing custom commands for servers administrators do some of this changes in-game, but only the safe changes without restart.

### DIFF
--- a/NachoSpigot-Server/src/main/java/dev/cobblesword/nachospigot/NachoConfig.java
+++ b/NachoSpigot-Server/src/main/java/dev/cobblesword/nachospigot/NachoConfig.java
@@ -16,4 +16,19 @@ public class NachoConfig {
     public int playerTimeStatisticsInterval = 20;
     public String serverBrandName = "NachoSpigot";
     public boolean enableAntiCrash = true;
+    public boolean infiniteWaterSources = true;
+    public boolean leavesDecayEvent = true;
+    public boolean enableMobAI = true;
+    public boolean enableMobSound = true;
+    public boolean enableEntityActivation = true;
+    public boolean enableLavaToCobblestone = true;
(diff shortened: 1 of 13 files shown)


Lucaskyy:
Awesome! Thank you so much for this :) <3

PR #104, opened 2021-07-19
rework/networking

Lucaskyy:
followup of #102 

reworked some of the networking, haven't tested it yet though.
some of the patches are from Minestom, possibly adding more.

added support for KQueue
made the packet buffer larger (going to add a config option for that)

and some other things :)

### DIFF
--- a/NachoSpigot-Server/src/main/java/org/bukkit/craftbukkit/Main.java
+++ b/NachoSpigot-Server/src/main/java/org/bukkit/craftbukkit/Main.java
@@ -20,18 +20,6 @@ public class Main {
     public static boolean useConsole = true;
 
     public static void main(String[] args) {
-//        [Nacho-0034]
-//        Remove this message from TacoSpigot, this won't work on Java 8 or higher because JAVA_1_8 doesn't exist on the JavaVersion enum when using these versions. (I have no clue why)
-//        So what was the point of making it support Java 14 if it never worked because of this ¯\_(ツ)_/¯
-//        Changed Java 8 message to Java 15.
-//        ---------------------------------------------------------------
-//        if (!SystemUtils.isJavaVersionAtLeast(JavaVersion.JAVA_1_8)) {
-//            System.err.println("TacoSpigot requires java 8");
-//            System.err.println("Oracle dropped all support for java " + SystemUtils.JAVA_VERSION);
-//            System.err.println("Please update to use TacoSpigot and the numerous bug-fixes, performance improvements, and security fixes");
-//            System.err.println("Shutting down");
-//            System.exit(1);
-//        }
         try {
             if(!SystemUtils.isJavaVersionAtLeast(JavaVersion.JAVA_15)) {
                 System.err.println("It seems like you are not using Java 15!");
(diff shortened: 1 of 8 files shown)


Lucaskyy:
@hpfxd Does this solve it?

CyberFlameGO (approved):

Lucaskyy:
FYI: This allows you to manually select a type of Epoll, KQueue, and NIO. If none are selected or the selected type is not supported it will utilize fall-through to select the one that's compatible with your system :)

Never knew fall-through could be handy, haha

hpfxd (approved):

Lucaskyy:
1 more commit coming soon, if gpg-agent decides to work lul

Lucaskyy:
there we go, does that look good to you now? @hpfxd

Lucaskyy:
it checks if `use-native-transport` is true and the server is running on Linux, if so just use Epoll if not use the `transport-to-use` variable

hpfxd:
If `use-native-transport` is true and the server is on a KQueue-supported platform, it should probably use it, not just for Epoll

Lucaskyy:
@hpfxd Done. I'll merge it now then :)

CyberFlameGO:
> 1 more commit coming soon, if gpg-agent decides to work lul, it's taking forever

ayyy finally someone else who signs their commits

NachoSpigot-Server/src/main/java/org/bukkit/craftbukkit/Main.java
```diff
@@ -20,18 +20,6 @@
...
-//        ---------------------------------------------------------------
-//        if (!SystemUtils.isJavaVersionAtLeast(JavaVersion.JAVA_1_8)) {
-//            System.err.println("TacoSpigot requires java 8");
-//            System.err.println("Oracle dropped all support for java " + SystemUtils.JAVA_VERSION);
-//            System.err.println("Please update to use TacoSpigot and the numerous bug-fixes, performance improvements, and security fixes");
-//            System.err.println("Shutting down");
-//            System.exit(1);
-//        }
         try {
             if(!SystemUtils.isJavaVersionAtLeast(JavaVersion.JAVA_15)) {
```
crafter23456, line 24: Should it be changed to java 16 because protocollib works fine?

Lucaskyy: We *could*, yes. But Java 15 is recommended either way.

bob7l:
Linux will dynamically adjust the SO_SNDBUF, so setting it shouldn't be needed (Or may even cause problems?). Might only be useful for OS's like Windows.

Lucaskyy:
Oh, didn't know about that. Should I only do that on Windows then? Since as far as I know, SO_SNDBUF tells Netty how big the buffer should be, which in turn allows for more packets to be sent each tick. A bigger Netty buffer should allow for higher play counts, right? At least, that's what I got told.

bob7l:
> 
> 
> Oh, didn't know about that. Should I only do that on Windows then? Since as far as I know, SO_SNDBUF tells Netty how big the buffer should be, which in turn allows for more packets to be sent each tick. A bigger Netty buffer should allow for higher play counts, right? At least, that's what I got told.

There's a min, default, and max size to that buffer on Linux. The max size is gigantic. It'll dynamically adjust it when it needs to (And if there's enough memory to do so...). Windows on the other hand is probably just a static value.

Lucaskyy:
Ah okay. So then, I guess that SO_SNDBUF is not the fix for the Netty buffer lag?

bob7l:
> 
> 
> Ah okay. So then, I guess that SO_SNDBUF is not the fix for the Netty buffer lag?

I doubt it'll effect it. The one confirmed thing this change WILL do however is increase memory usage

Lucaskyy:
Ah okay, thanks for telling me!

PR #106, opened 2021-07-19
serverName & serverVersion fit config

crafter23456:
https://imgur.com/a/Vvnyakb
It get the serverName from the config. I think its only the first part of "This server is running NachoSpigot version" which will get customized.
It changes the serverVersion that the version don't look like "1.8.8", it look like "NachoSpigot-1.8.8".  Paper 1.17 has for example "git-Paper-114". 
The whole commit should only change the server log message & the timings. It doesnt affect the MOD in F3. If you want that it say everytime NachoSpigot in the log, you can simply change it back :P
Change readme from java 15 to java 16 because it works know, don't know why it got changed. If it's wrong you know what to do.


Lucaskyy:
We can't do that because loading Nacho this early might cause issues if I recall correctly. Also, the Java version change has been reverted in the network rework PR.

PR #107, opened 2021-07-19
[Flamepaper 0113] Check channel before reading

crafter23456:


Lucaskyy:
Thank you! :) I've added the change, and I will close this PR because you accidentally added the other commits.

PR #108, opened 2021-07-19
Async TNT & Some other optimizations

Lucaskyy:
Added async TNT by asyncing the rayTrace method which is called by the getBlockDensity method.

Also added some other optimizations and QoL updates.

I was able to achieve little to no change in TPS using this async TNT method, I really like it.


Lucaskyy:
Also, yes I've tested everything and it didn't cause any issues.

PR #116, opened 2021-07-20
Added disableSpongeAbsorption settings.

benfiratkaya:
As in version 1.7, I made the sponge's water absorption adjustable from the settings. By default this setting is false.
I use this frequently.


Lucaskyy:
Added your changes into a mapped file from BuildTools.
Thank you for this PR! :)

PR #126, opened 2021-07-20
feat(performance): async and optimize some things

Lucaskyy:
Asyncing and optimizing some things, yeah that.

- [ ] Async Entity Tracker
- [ ] Try to minimize the usage of BlockPosition
- [ ] Possibly switch to zlib-ng?
- [ ] Optimize packet decoding by using Suppliers

Possibly more to come.


NachoSpigot-Server/src/main/java/net/minecraft/server/EntityTracker.java
```diff
@@ -185,6 +188,27 @@ public void updatePlayers() {
                 }
             }
         }
+//        int offset = 0;
+//        CountDownLatch latch = new CountDownLatch(4);
+//        for (int i = 1; i <= 4; i++) {
+//            int localOffset = offset++;
+//            Runnable runnable = () -> {
+//                for (int in = localOffset; in < this.c.size(); in += 4) {
+//                    this.c.get(in).a(); // TODO this is not the proper call! make sure to fix this.
```
Lucaskyy, line 197: This needs fixing ^

Ytnoos:
I would suggest to add a new task: converting all the packet's classes of the registry to Suppliers, replacing newInstance() calls with a faster "new" for every packet received. Velocity already does it, so you can take a look at it.

Lucaskyy:
@Ytnoos That belongs to the last task :) Hence I was working on remapping EnumProtocol to get an understanding of what does what. I'll edit it to make it clear.

NachoSpigot-Server/src/main/java/net/minecraft/server/EntityTracker.java
Lucaskyy: Lew has made a patch for this, so it should be fixed soon :)

LewUwU:
> I would suggest to add a new task: converting all the packet's classes of the registry to Suppliers, replacing newInstance() calls with a faster "new" for every packet received. Velocity already does it, so you can take a look at it.

 Optimize packet decoding by using Suppliers

andreasdc:
When release? 🗡️

LewUwU:
> When release?

I will release the Async Entity Tracker when i finish somethings on my own fork (I have customers too)

andreasdc:
@LewUwU what fork do you have?

LewUwU:
> @LewUwU what fork do you have?

InsanePaper

andreasdc:
> > @LewUwU what fork do you have?
> 
> InsanePaper

That's cool, don't you think that async movement and hits would be a better idea?

LewUwU:
> > > @LewUwU what fork do you have?
> > 
> > 
> > InsanePaper
> 
> That's cool, don't you think that async movement and hits would be a better idea?
> When I look at this I think that's good idea. :D
> image

Not really, the performance depends on your server hardware, moving something to another threada will be better for the main but the thread will be fucked anyways, better will be to fix the lag root

andreasdc:
Could be, so movement is still at the top of profilers.

Rastrian:
Any update about the design system of async? i was thinking in implement async hit detection and knockback after this PR be released, any progress into this?

HeathLoganCampbell:
Be careful with async entity trackers, they are known to have some weird edge cases

andreasdc:
> Be careful with async entity trackers, they are known to have some weird edge cases

I'm testing and experiencing some currently haha :D

HeathLoganCampbell:
Sadly you can't touch BlockPosition too without breaking a bunch of plugins

LewUwU:
> Sadly you can't touch BlockPosition too without breaking a bunch of plugins

Not really, since you can do methods with int3 and using the old method that uses BlockPosition with a return

public boolean method(int int int) {
}

public boolean method(position) {
return method(position.x, position.y, position.z);
}

andreasdc:
I have improved packets processing and everything is really smooth, I notice how it is so slow normaly.
If you don't want to share and do async yet, maybe you could do some kind of connected flushes, like these from entity tracker?

PR #129, opened 2021-07-23
Heavily optimize TacoSpigot HopperPusher

LewUwU:


NachoSpigot-Server/src/main/java/net/techcable/tacospigot/HopperPusher.java
```diff
@@ -8,29 +8,13 @@
...
-                     * We add one to getY(), just like {@link TileEntityHopper#b(IHopper)}
-                     */
-                    AxisAlignedBB hopperBoundingBox = hopper.getHopperLookupBoundingBox();
-                    if (hopperBoundingBox.b(this.getBoundingBox())) { // AxisAlignedBB.b(AxisAlignedBB) -> isIntersect()
-                        return hopper;
-                    }
-                }
+        int y1 = MathHelper.floor(getY());
+        for (int y = y1; y > y1 - 2; y--) {
+            TileEntityHopper hopper = HopperHelper.getHopper(getWorld(), new BlockPosition(getX(), y, getZ()));
```
Lucaskyy, line 15: Would it help if we have getX() and getZ() as a local variable? Because now we're continuously calling getX() and getZ() which then calls another method to eventually get the variable address. Not sure if the java compiler fixes/inlines this though?

LewUwU: maybe

PR #134, opened 2021-07-27
Update /version

Elierrr:
/version currently doesn't work since it points to TacoSpigot instead of NachoSpigot.  
Useful for issues


Lucaskyy:
Thank you so much for fixing this! <3 :)

PR #142, opened 2021-08-07
lastDamageEvent isn't useless.

AhmHkn0:
Skript's "on death" code uses lastDamageEvent. So we must take lastDamageEvent


Lucaskyy:
I forgot about that yeah, had to uncomment that. Thanks for this PR!

Issue #152, opened 2021-08-31
Add missing options to config


andreasdc:
Idea: When loading config with missing options add new options to it

Rastrian:
any example?

andreasdc:
> any example?

I'm talking about adding options to file when they are not specified

PR #154, opened 2021-09-01
Fix https://github.com/CobbleSword/NachoSpigot/issues/152

Elierrr:


andreasdc:
When called after loading config it doesn't set it to default

Lucaskyy:
Alright then, thank you for this PR!

PR #159, opened 2021-09-13
docs: preparations for native java 17 support

CyberFlameGO:
[ ] Verify compatibility
[ ] Utilize any new functionality if present

(Closes #158)

PR #160, opened 2021-09-13
Dependency updates (needs compatibility validation

CyberFlameGO:


NachoSpigot-API/pom.xml
```diff
@@ -10,29 +10,19 @@
...
     <name>NachoSpigot-API</name>
     <description>An enhanced plugin API for Minecraft servers.</description>
 
     <properties>
-        <!--PaperSpigot - Bump to 1.8 - This will haunt me -->
         <additionalparam>-Xdoclint:none</additionalparam>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>15</maven.compiler.source>
+        <maven.compiler.target>15</maven.compiler.target>
```
Elierrr, line 22: https://github.com/CobbleSword/NachoSpigot/pull/159#discussion_r707199048

NachoSpigot-API/pom.xml
```diff
@@ -10,29 +10,19 @@
     </parent>
 
     <artifactId>api</artifactId>
-    <version>1.8.8-R0.2-SNAPSHOT</version>
+    <version>1.8.8-R0.3-SNAPSHOT</version>
```
Elierrr, line 13: Why the change?

Elierrr (changes requested):
Everything else seems fine.

NachoSpigot-API/pom.xml
CyberFlameGO: That's actually the latest 1.8.8 release, I'm pretty sure putting it as `R0.2` was an accident

CyberFlameGO:
https://github.com/CobbleSword/NachoSpigot/pull/160#discussion_r707219371 fixed this just before you sent it :)

Again, for the force-pushes, they were to amend commit messages (but this time not signing)

NachoSpigot-API/pom.xml
Elierrr: You only changed the server back to Java 8 and now the server won't compile :p

CyberFlameGO: Gadzooks! I'll fix that now haha, I must've missed it

CyberFlameGO:
Another force-push which was necessary to amend a commit, and another regular commit was added as per https://github.com/CobbleSword/NachoSpigot/pull/160#discussion_r707234069

Elierrr:
LGTM, I'll test it when I can.

andreasdc:
What is the difference between R0.3 and R0.2?

NachoSpigot-API/pom.xml
Elierrr: Spigot and Paper both have RO.1 but TacoSpigot [changed it](https://github.com/TacoSpigot/TacoSpigot/commit/816cde55dfc03d813ca55e0d66a49eb1fae11f21#diff-3e189d8b27d4e715e93f4f3e6f2875558d4f3f01d5a4526c9b7a1eed3dff3d54) to RO.2. I have no idea how the versioning works but I think it should be RO.1

CyberFlameGO:
I could've sworn I've seen a R0.3 before; apparently not then. My bad

Lucaskyy:
I've spoken with @LewUwU before about this, this sadly breaks a lot of plugins. SnakeYAML will completely stop working for nearly every plugin on the latest version, due to strict security measures in the new version which is a pain to fix. And the MySQL Connector also shouldn't be updated since that also breaks a lot of plugins. As long as the dependency is internal you can update it without any issues, but as for external dependencies please watch out. And indeed, in my opinion, it's better to leave it R0.2 for compatibility reasons and in case something stupid happens.

andreasdc:
> n

What's the difference betwen r3 and r2?

LewUwU (changes requested):
R0.3 -> R0.2, you're breaking tons of things

NachoSpigot-API/pom.xml
LewUwU: Instead of dropping java 8 without any reason, do a issue about why nachospigot should only support java 15-17

NachoSpigot-API/pom.xml
LewUwU: The sub-release of a version

NachoSpigot-API/pom.xml
Elierrr: It was on accident and already corrected.

NachoSpigot-API/pom.xml
Elierrr: Then it should be RO.1

andreasdc:
> R0.3 -> R0.2, you're breaking tons of things

like?

Lucaskyy:
NMS plugins mostly.

andreasdc:
But package is currently v1_8_R3

Lucaskyy:
I understand, but a lot of plugins look at the version string v1_8_R3 and determine that way that a server is running 1.8. Changing that breaks that. Don't do that.

andreasdc:
> I understand, but a lot of plugins look at the version string v1_8_R3 and determine that way that a server is running 1.8. Changing that breaks that. Don't do that.

You could always recode these plugins

Lucaskyy:
What the heck? No? Me changing a tiny number requires me to recode tons of plugins? No thanks.

andreasdc:
> What the heck? No? Me changing a tiny number requires me to recode tons of plugins? No thanks.

People that want R4 would have to recode, that's not big problem. I personally use only plugins coded by myself.
What changing from r3 to r4 brings?

Lucaskyy:
It has nothing to do with compiling.
Some plugins read the package name (net.minecraft.v1_8_R3) and changing it will break a lot. Not happening. End of discussion.

andreasdc:
> .

Yea but does it bring any improvement?

Lucaskyy:
R3 to R4? No, absolutely not lol. All it does it break things

andreasdc:
All right, I get it. Are changes viewable somewhere maybe?

Lucaskyy:
Which changes?

andreasdc:
From R3 to R4

Lucaskyy:
Well the version was never supposed to be bumped anyway or whatever happened to it. So, just ignore it.

CyberFlameGO:
I don't actually think I touched what you're referring to. I changed some
XML tag at the top, not the NMS package

CyberFlameGO:
If that's what you mean, you can fix it yourself because I've given
maintainers permission to act on the PR

CyberFlameGO:
> I've spoken with @LewUwU before about this, this sadly breaks a lot of plugins. SnakeYAML will completely stop working for nearly every plugin on the latest version, due to strict security measures in the new version which is a pain to fix. And the MySQL Connector also shouldn't be updated since that also breaks a lot of plugins. As long as the dependency is internal you can update it without any issues, but as for external dependencies please watch out. And indeed, in my opinion, it's better to leave it R0.2 for compatibility reasons and in case something stupid happens.

Fixed although I changed us to R0.1 in accordance with Spigot/Paper upstream, as apparently TacoSpigot changed it

CyberFlameGO:
https://github.com/CobbleSword/NachoSpigot/pull/160#discussion_r707560120 will do :)

NachoSpigot-Server/pom.xml
```diff
@@ -60,12 +58,12 @@
         <dependency>
             <groupId>it.unimi.dsi</groupId>
             <artifactId>fastutil</artifactId>
-            <version>8.5.4</version>
+            <version>8.5.5</version>
```
Elierrr, line 61:
```suggestion
            <version>8.5.6</version>
```

NachoSpigot-Server/pom.xml
```diff
@@ -245,7 +243,7 @@
                                 </relocation>
                                 <relocation>
                                     <pattern>net.minecraft.server</pattern>
-                                    <shadedPattern>net.minecraft.server.v${minecraft_version}</shadedPattern>
+                                    <shadedPattern>net.minecraft.server.v1_8_R3</shadedPattern>
```
Elierrr, line 246: On line 231
```patch
-                                    <shadedPattern>org.bukkit.craftbukkit.v${minecraft_version}</shadedPattern>
+                                    <shadedPattern>org.bukkit.craftbukkit.v1_8_R3</shadedPattern>
```

NachoSpigot-Server/pom.xml
Elierrr, line 61:
```suggestion
            <version>8.5.6</version>
```

Elierrr (changes requested):

NachoSpigot-Server/pom.xml
CyberFlameGO: what's this? i just removed the variable at the top and put its content where it was used, which is there

Elierrr: I know, but you forgot to change it for craftbukkit's relocation.

PR #168, opened 2021-10-03
Fix compiling on Java 17

Elierrr:


Lucaskyy:
Alright then, thanks for this PR! :)

PR #183, opened 2021-10-11
Use FastMath

ghost:
Originally from 44b3807e1fb903b83aabfbbcc0c9440953f1c5c3
Also updates fast-uuid to version 0.2.0


andreasdc:
MathHelper is better by 40-100%, just checked

andreasdc:
> you should pr it over then :)

maybe someday I will :)

vocant:
did you know we're in 2021? jafama fastmath offers better performance only on old processors and java 1.6.

Elierrr:
> did you know we're in 2021? jafama fastmath offers better performance only on old processors and java 1.6.

Doesn't make it worse than code from 2014.

PR #190, opened 2021-10-15
Revert using fastmath while retaining the fast-uuid update, and remove useless repo

CyberFlameGO:
This is being reverted because of the comments on the (squashed) commit, and since it was squashed I can't revert only the FastMath part.
@Izibane @EthsMC @MWHunter @Elierrr @Someonewhoisaperson @FurmigaHumana

Supersedes #189 and #187

### DIFF
--- a/NachoSpigot-Server/pom.xml
+++ b/NachoSpigot-Server/pom.xml
@@ -117,17 +117,10 @@
         <dependency>
             <groupId>com.eatthepath</groupId>
             <artifactId>fast-uuid</artifactId>
-            <version>0.1</version>
+            <version>0.2.0</version>
         </dependency>
     </dependencies>
 
-    <repositories>
-        <repository>
-            <id>elmakers-repo</id>
-            <url>https://maven.elmakers.com/repository/</url>
-        </repository>
-    </repositories>
-  
     <!-- This builds a completely 'ready to start' jar with all dependencies inside -->
     <build>
         <finalName>NachoSpigot-Server</finalName>


CyberFlameGO:
Not explicitly, but take a look at the comments of https://github.com/CobbleSword/NachoSpigot/commit/dd5ff5959a65ac74504a0f90a2d37503f408fdf6

andreasdc:
"Jafama (Java Fast Math) is a Java library aiming at providing faster versions of java.lang.Math treatments, at the eventual cost of 1e-15ish accuracy errors"
So I don't think that would impact "3.00000001 reach" or "0.00000001 blocks of movement"

Lucaskyy:
Thanks Hunter for showing this! I won't use FastMath then, and I mean, is FastMath currently that much faster than Java 17's math?

andreasdc:
> Yes, but anticheats often use the server sided trig table as the client and server use the same table. Changing this table could affect the ability for anticheats to correctly predict the player's movement.

So you block players who use fast math client-side?

MWHunter:
No, you usually also include the fastmath trig table in the anticheat and switch between them when needed because sp614x blames the anticheats for being bad since they are too precise, and refuses to remove fastmath.  Changing the server's trig table will break anticheats that rely on the server sided trig table being the same as the client sided one.

andreasdc:
> Yes, and that's enough to flag a user for cheating, on default config, at high speeds: https://www.youtube.com/watch?v=zgGjIR8jyoU

You have shown usage of sin and cos, not other methods

andreasdc:
> > > Yes, and that's enough to flag a user for cheating, on default config, at high speeds: https://www.youtube.com/watch?v=zgGjIR8jyoU
> > 
> > 
> > You have shown usage of sin and cos, not other methods
> 
> That's because those are the only methods that produce incorrect results.

Why don't keep other methods then? :D

Izibane:
Which are the methods supposedly meant to give a performance boost with new
tables. Even though the developers cant name a single cpu it improves
performance on that you can buy.

Elierrr:
> > > > Yes, and that's enough to flag a user for cheating, on default config, at high speeds: https://www.youtube.com/watch?v=zgGjIR8jyoU
> > > 
> > > 
> > > You have shown usage of sin and cos, not other methods
> > 
> > 
> > That's because those are the only methods that produce incorrect results.
> 
> Why don't keep other methods then? :D

I did that in #187 but this was merged instead.

Rastrian:
I still think the usage of FastMath helps into server-side.

"bruh but breaks anticheats bc of the number disparity"

Okay, just add as an config, doesnt need remove from the entire project, this can be useful. Precision sometimes come it a cost? yes, but not everybody needs precision for everything.

Anticheat developers who trust in the core should at least preview this, using their own methods to simulate vanilla behavior.

Im gonna make a PR with optional FastMath config.

MWHunter:
No performance improvement from fastmath... only wasted developer time.

Rastrian:
> No performance improvement from fastmath... only wasted developer time.

Say for yourself, i already had seen great anticheats who supports FastMath without any mistake, if you disagree, will be a config option just for this.

andreasdc:
> > No performance improvement from fastmath... only wasted developer time.
> 
> Say for yourself, i already had seen great anticheats who supports FastMath without any mistake, if you disagree, will be a config option just for this.

How to precisely compare 2 methods? I tried to do that but it was pretty random.

Rastrian:
> > > No performance improvement from fastmath... only wasted developer time.
> > 
> > 
> > Say for yourself, i already had seen great anticheats who supports FastMath without any mistake, if you disagree, will be a config option just for this.
> 
> How to precisely compare 2 methods? I tried to do that but it was pretty random.

Being honest, i used FasthMath only for PvP advantages, most PvP servers uses similar implementation, during async process (hit detection async, knockback also), this can handle more people using the same actions simultaneously.

This usage is one of distinct samples, you can  get an more expressive result with high player counts doing the same function simultaneously (300+ as far i tested on the same instance).

andreasdc:
> > > > No performance improvement from fastmath... only wasted developer time.
> > > 
> > > 
> > > Say for yourself, i already had seen great anticheats who supports FastMath without any mistake, if you disagree, will be a config option just for this.
> > 
> > 
> > How to precisely compare 2 methods? I tried to do that but it was pretty random.
> 
> Being honest, i used FasthMath only for PvP advantages, most PvP servers uses similar implementation, during async process (hit detection async, knockback also), this can handle more people using the same actions simultaneously.
> 
> This usage is one of distinct samples, you can get an more expressive result with high player counts doing the same function simultaneously (300+ as far i tested on the same instance). It can affect other related functions, but need a expressive test, adding the option we can threat this effectively with the time, and will be optional, defaultly disabled.
> 
> When i finish the PR, some ticking fixes (legacy code from minecraft server, spigot, paper) will be fixed, you maybe will notice some improvement.

I did many fixes by myself too, maybe you will touch what I did. I mean how to compare performance of Math and FastMath precisely?

Rastrian:
@andreasdc as far I seen, you can't precisely.

But jafama library doesn't just have these methods, also we got methods who make the math more quickier than implemented now, but not precisely good.
image
the idea is just provide an good approximately result instead of using floating point numbers as crazy.

andreasdc:
> @andreasdc as far I seen, you can't precisely.
> 
> But jafama library doesn't just have these methods, also we got methods who make the math more quickier than implemented now, but not precisely good. image the idea is just provide an good approximately result instead of using floating point numbers as crazy.

I think there should be way to compare performance of it.

Rastrian:
> > @andreasdc as far I seen, you can't precisely.
> > But jafama library doesn't just have these methods, also we got methods who make the math more quickier than implemented now, but not precisely good. image the idea is just provide an good approximately result instead of using floating point numbers as crazy.
> 
> I think there should be way to compare performance of it.

You would need do individual comparisons between this. Quick methods as I mentioned, are extremely faster and had an notable difference, but at knockback for example, if i trigger an quick method, the disparity can be huge,

PR #195, opened 2021-10-17
may fix #194

### LINKED ISSUE #194: NachoSpigot doesn't load on Windows
Works fine on Arch Linux and WSL

`java.lang.UnsatisfiedLinkError: Could not load library. Reasons: [no jansi64-git-NachoSpigot-"e5a4b96" in java.library.path, no jansi-git-NachoSpigot-"e5a4b96" in java.library.path, no jansi in java.library.path, The filename, directory name, or volume label syntax is incorrect]
        at org.fusesource.hawtjni.runtime.Library.doLoad(Library.java:182)
        at org.fusesource.hawtjni.runtime.Library.load(Library.java:140)
        at org.fusesource.jansi.internal.CLibrary.<clinit>(CLibrary.java:42)
        at org.fusesource.jansi.AnsiConsole.wrapOutputStream(AnsiConsole.java:48)
        at org.fusesource.jansi.AnsiConsole.<clinit>(AnsiConsole.java:38)
        at org.bukkit.craftbukkit.Main.main(Main.java:191)`

CyberFlameGO:
fixes gitignore too but that's nothing important

### DIFF
--- a/NachoSpigot-API/.gitignore
+++ b/NachoSpigot-API/.gitignore
@@ -21,6 +21,8 @@
 /dist
 /manifest.mf
 
+# Mac filesystem dust
+.DS_Store
 
 # intellij
 *.iml
--- a/NachoSpigot-Server/.gitignore
+++ b/NachoSpigot-Server/.gitignore
@@ -25,6 +25,9 @@ dependency-reduced-pom.xml
 
 /world
 
+# Mac filesystem dust
+.DS_Store
+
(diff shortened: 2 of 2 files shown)


CyberFlameGO:
Basically I just added a dependency in the hopes of it working because I saw it as a fix on StackOverflow, and it seemed logical to me. I might have the wrong version though but we'll see if this is a fix

PR #212, opened 2021-10-29
Add EntityHider (for Practice and MiniGames servers)

SaithTime:
Added an entity hider system if the shooter is not visible by other.

Method based on a private patched version of SpigotX


Elierrr (changes requested):
Add a configuration option.

pom.xml
```diff
@@ -22,6 +22,10 @@
         <module>NachoSpigot-API</module>
     </modules>
 
+    <build>
```
CyberFlameGO, line 25: Not everyone needs the API built, so in my opinion this is pretty pointless

CyberFlameGO (changes requested):
I agree with Elier, and it's easy to make this a configuration option

pom.xml
CyberFlameGO: Nevermind iirc the server needs the api to build upon

CyberFlameGO:
You could also add `clean` (**NOT REPLACE**, so just make it `clean install` instead of just `install`) to the [server module](https://github.com/CobbleSword/NachoSpigot/blob/master/NachoSpigot-Server/pom.xml) default goals, and fix the [README.md](https://github.com/CobbleSword/NachoSpigot/blob/master/README.md) instructions. You may even be able to just add the default goals to the [parent POM](https://github.com/CobbleSword/NachoSpigot/blob/master/README.md) instead of messing around with each module

Lucaskyy:
Superseded by #213

PR #215, opened 2021-11-02
api to hide player in tab

SaithTime:
allows you to choose whether or not to hide a player in the tab-list with the hidePlayer() method.


Elierrr (approved):

sadcenter (approved):
I don't think this is very needed, but if this will be useful for someone why not?

SaithTime:
This can be useful for practice servers (used as a config on minemen.club for example) but it could also be used on minigames servers to leave players in the tab when hiding players in a lobby.
It's small but it's really practical

xzolw (approved):

Lucaskyy:
Thank you for this PR! :)

PR #225, opened 2021-11-11
Fixes the null pointer due to bukkit logger use. 

sadcenter:
The server isn't yet initialized when NachoConfig uses bukkit logger.


Elierrr (approved):
LGTM! You should also revert https://github.com/CobbleSword/NachoSpigot/commit/d80035ca9d41fb78a187a04affc7edc46ca712c8 as it's unneeded here.

Lucaskyy:
Thanks for this PR! :)

PR #231, opened 2021-11-13
Ticking fixes, tile optimization, optional fast math, chunk tick optimization...

Rastrian:
- Added toggleable fast math config (default disabled).
- Added toggleable [infinisleeper thread.](https://web.archive.org/web/20150924165427/https://blogs.oracle.com/dholmes/entry/inside_the_hotspot_vm_clocks).
- Optimized tile entities.
- [EntityLiving container inventory ticking optimized - ported from Spigot.](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/spigot/browse/CraftBukkit-Patches/0131-Use-FastMatches-for-ItemStack-Dirty-Check.patch?at=a22b705b13d5141e13896333d690a81985ea2e74).
- Fixed ticking issues, some legacy code/solutions had been disabled and replaced for a more appropriate solution.
- Added [TCP_FASTOPEN.](https://en.wikipedia.org/wiki/TCP_Fast_Open).
- [Don't tick blocks if neighbor chunks aren't loaded.](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/spigot/browse/CraftBukkit-Patches/0137-Don-t-tick-blocks-if-neighbor-chunks-aren-t-loaded.patch?at=a22b705b13d5141e13896333d690a81985ea2e74).

You can get a built JAR here: https://github.com/Argarian-Network/NachoSpigot/actions
Tests and suggestions are appreciated.


NachoSpigot-Server/src/main/java/net/minecraft/server/EntityPlayer.java
```diff
@@ -208,9 +208,9 @@ public void t_() {
         }
         
         // PaperSpigot start - Configurable container update tick rate
-        if (--containerUpdateDelay <= 0) {
+        if (/*--containerUpdateDelay <= 0*/ true) {
```
andreasdc, line 211: "Useless"?

NachoSpigot-Server/src/main/java/net/minecraft/server/DedicatedServer.java
```diff
@@ -42,22 +42,24 @@
     public DedicatedServer(joptsimple.OptionSet options) {
         super(options, Proxy.NO_PROXY, DedicatedServer.a);
         // CraftBukkit end
-        Thread thread = new Thread("Server Infinisleeper") {
-            {
-                this.setDaemon(true);
-                this.start();
-            }
+        if (!NachoConfig.disableInfinitSleeperThreadUsage) {
```
andreasdc, line 45: What is that thread for?

andreasdc:
It would be nice to see performance comparisons between old methods and FastMath.

NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
```diff
@@ -322,4 +322,18 @@ private static void disableFallAnimation() {
...
+    private static void disableInfinitSleeperThreadUsage() {
+        disabledFallBlockAnimation = getBoolean("settings.disable-infinisleeper-thread-usage", false);
+        c.addComment("settings.disable-infinisleeper-thread-usage", "Disable infinisleeper thread usage, just enable this if you know what are you doing.");
+    }
+
+    public static boolean enableFastMath;
+
+    private static void enableFastMath() {
+        disabledFallBlockAnimation = getBoolean("settings.enable-fastmath", false);
+        c.addComment("settings.enable-fastmath", "Enable Fast Math usage, this sometimes break anticheats, becareful.");
```
tomasalves8, line 336: Typo: "be careful"

NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
```diff
@@ -322,4 +322,18 @@ private static void disableFallAnimation() {
         c.addComment("settings.disabled-block-fall-animation", "Disables the fall animation for blocks");
     }
 
+    public static boolean disableInfinitSleeperThreadUsage;
+
+    private static void disableInfinitSleeperThreadUsage() {
+        disabledFallBlockAnimation = getBoolean("settings.disable-infinisleeper-thread-usage", false);
```
Elierrr, line 328: Bump config version in the init method

NachoSpigot-Server/src/main/java/net/minecraft/server/MathHelper.java
```diff
@@ -10,29 +12,32 @@
     private static final double d;
     private static final double[] e;
     private static final double[] f;
+    private static final boolean fastMathMode = NachoConfig.enableFastMath;
 
     public static float sin(float var0) {
-        return b[(int)(var0 * 10430.378F) & '\uffff'];
+        return (fastMathMode == true ? ((float) FastMath.sin(var0)) : (b[(int)(var0 * 10430.378F) & '\uffff']));
```
Elierrr, line 18:
```suggestion
        return b[(int)(var0 * 10430.378F) & '\uffff'];
```

This not only would cause anti-cheat issues but also probably issues with the server. This will also let us remove the caution for Fast math, as these are the only methods that are different in my tests.

NachoSpigot-Server/src/main/java/net/minecraft/server/MathHelper.java
```diff
@@ -10,29 +12,32 @@
...
     }
 
     public static float cos(float var0) {
-        return b[(int)(var0 * 10430.378F + 16384.0F) & '\uffff'];
+        return (fastMathMode == true ? ((float) FastMath.cos(var0)) : (b[(int)(var0 * 10430.378F + 16384.0F) & '\uffff']));
```
Elierrr, line 22:
```suggestion
        return b[(int)(var0 * 10430.378F + 16384.0F) & '\uffff'];
```

Same as above

NachoSpigot-Server/src/main/java/net/minecraft/server/MathHelper.java
```diff
@@ -295,9 +306,9 @@ public static double i(double var0) {
         f = new double[257];
 
         for(var0 = 0; var0 < 257; ++var0) {
-            double var1 = (double)var0 / 256.0D;
-            double var3 = Math.asin(var1);
-            f[var0] = Math.cos(var3);
+            double var1 = (double) var0 / 256.0D;
+            double var3 = (fastMathMode == true ? (FastMath.asin(var1)) : (Math.asin(var1)));
+            f[var0] = (fastMathMode == true ? (FastMath.cos(var3)) : (Math.cos(var3)));
```
Elierrr, line 311:
```suggestion
            f[var0] = Math.cos(var3);
```

Same as above

NachoSpigot-Server/src/main/java/net/minecraft/server/MathHelper.java
```diff
@@ -286,7 +293,11 @@ public static double i(double var0) {
     static {
         int var0;
         for(var0 = 0; var0 < 65536; ++var0) {
-            b[var0] = (float)Math.sin((double)var0 * 3.141592653589793D * 2.0D / 65536.0D);
+            if (fastMathMode) {
+                b[var0] = (float) FastMath.sin((double) var0 * FastMath.PI * 2.0D / 65536.0D);
+            } else {
+                b[var0] = (float) Math.sin((double) var0 * 3.141592653589793D * 2.0D / 65536.0D);
```
Elierrr, line 299: Same as above

Elierrr (changes requested):

NachoSpigot-Server/src/main/java/net/minecraft/server/MathHelper.java
andreasdc: You should do other tests then.

NachoSpigot-Server/src/main/java/net/minecraft/server/World.java
```diff
@@ -64,7 +66,7 @@ private void guard()
     // Spigot end
     protected final Set<Entity> g = Sets.newHashSet(); // Paper
     //public final List<TileEntity> h = Lists.newArrayList(); // PaperSpigot - Remove unused list
-    public final List<TileEntity> tileEntityList = Lists.newArrayList();
+    public final me.rastrian.dev.OptimizedWorldTileEntitySet tileEntityList = new me.rastrian.dev.OptimizedWorldTileEntitySet();
```
Lucaskyy, line 69: Why qualify the entire classpath & name when you're importing the class above?

NachoSpigot-Server/src/main/java/net/minecraft/server/MobSpawnerAbstract.java
```diff
@@ -23,16 +23,16 @@
     private MobSpawnerAbstract.a spawnData;
     private double e;
     private double f;
-    private int minSpawnDelay = 200;
-    private int maxSpawnDelay = 800;
+    private int minSpawnDelay = 200 / 20;
+    private int maxSpawnDelay = 800 / 20;
```
Lucaskyy, line 27: What's with the `/ 20`?

NachoSpigot-Server/src/main/java/net/minecraft/server/MobSpawnerAbstract.java
```diff
@@ -23,16 +23,16 @@
...
     private int spawnCount = 4;
     private Entity j;
     private int maxNearbyEntities = 6;
     private int requiredPlayerRange = 16;
     private int spawnRange = 4;
     private int tickDelay = 0; // PaperSpigot
 
-    static final int FAIL_DELAY_TICKS = 100; // Mobspawner Fail Delay
+    static final int FAIL_DELAY_TICKS = 100 / 20; // Mobspawner Fail Delay
```
Lucaskyy, line 35: Same here

NachoSpigot-Server/src/main/java/net/minecraft/server/MobSpawnerAbstract.java
```diff
@@ -294,9 +294,9 @@ public void b(NBTTagCompound nbttagcompound) {
 
         if (!UtilColor.b(s)) {
             nbttagcompound.setString("EntityId", s);
-            nbttagcompound.setShort("Delay", (short) this.spawnDelay);
-            nbttagcompound.setShort("MinSpawnDelay", (short) this.minSpawnDelay);
-            nbttagcompound.setShort("MaxSpawnDelay", (short) this.maxSpawnDelay);
+            nbttagcompound.setShort("Delay", (short) (this.spawnDelay * 20));
+            nbttagcompound.setShort("MinSpawnDelay", (short) (this.minSpawnDelay * 20));
+            nbttagcompound.setShort("MaxSpawnDelay", (short) (this.maxSpawnDelay * 20));
```
Lucaskyy, line 299: Same here

NachoSpigot-Server/src/main/java/org/bukkit/craftbukkit/block/CraftCreatureSpawner.java
```diff
@@ -69,11 +69,11 @@ public void setCreatureTypeByName(String creatureType) {
     }
 
     public int getDelay() {
-        return spawner.getSpawner().spawnDelay;
+        return spawner.getSpawner().spawnDelay * 20;
     }
 
     public void setDelay(int delay) {
-        spawner.getSpawner().spawnDelay = delay;
+        spawner.getSpawner().spawnDelay = delay / 20;
```
Lucaskyy, line 76: Same here

NachoSpigot-Server/src/main/java/org/bukkit/craftbukkit/block/CraftCreatureSpawner.java
```diff
@@ -69,11 +69,11 @@ public void setCreatureTypeByName(String creatureType) {
     }
 
     public int getDelay() {
-        return spawner.getSpawner().spawnDelay;
+        return spawner.getSpawner().spawnDelay * 20;
```
Lucaskyy, line 72: Same here

Lucaskyy (commented):
Looks good!

NachoSpigot-Server/src/main/java/net/minecraft/server/DedicatedServer.java
Lucaskyy: Here: https://web.archive.org/web/20150924165427/https://blogs.oracle.com/dholmes/entry/inside_the_hotspot_vm_clocks

Didn't read everything, but if I understand correctly when you do `Thread.sleep` the JVM checks every 10ms if it needs to stop suspending the thread. When you do something like this (called an infinisleeper) the JVM will instead check every millisecond if it needs to stop suspending.

I think I understand why they did this because the server sleeps when it's not processing a tick. So if the server doesn't need to do a lot, it will sleep most of the time and only tick the server 20 times per second. Hence. when you make a spark report, you see Thread.sleep with 95% usage (for me) which lets you know that the server didn't need to do a lot. If you have 300 players online, this will be a LOT lower since each tick takes longer, so less time to sleep.

But, let's say, a tick just got processed. The server now needs to sleep till the next tick. You want ticks to ALWAYS be 20 per second, and it must be synced with the time. But without an infinisleeper, when you do Thread.sleep to wait for the next tick, it can take up to 10ms longer for the thread to come back to life. This causes desyncs, which you don't want.

That's why I also don't really understand why this is a thing now, but I guess that customizability isn't a bad thing 🤷‍♂️

NachoSpigot-Server/src/main/java/org/bukkit/craftbukkit/block/CraftCreatureSpawner.java
sadcenter: @Lucaskyy ticks maybe? idk

Lucaskyy: Ah, I see.

sadcenter:
I think someone should test these changes in-game cause it's really massive PR

NachoSpigot-Server/src/main/java/net/minecraft/server/MathHelper.java
```diff
@@ -259,11 +266,11 @@ public static double b(double var0, double var2) {
             double var22 = (6.0D + var20 * var20) * var20 * 0.16666666666666666D;
             double var24 = var14 + var22;
             if (var8) {
-                var24 = 1.5707963267948966D - var24;
+                var24 = (fastMathMode == true ? (FastMath.PI / 2) : (1.5707963267948966D)) - var24;
```
Elierrr, line 269:
```suggestion
                var24 = Math.PI / 2 - var24;
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MathHelper.java
```diff
@@ -259,11 +266,11 @@ public static double b(double var0, double var2) {
...
             }
 
             if (var7) {
-                var24 = 3.141592653589793D - var24;
+                var24 = (fastMathMode == true ? (FastMath.PI) : (3.141592653589793D)) - var24;
```
Elierrr, line 273:
```suggestion
                var24 = Math.PI - var24;
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MathHelper.java
```diff
@@ -295,9 +306,9 @@ public static double i(double var0) {
         f = new double[257];
 
         for(var0 = 0; var0 < 257; ++var0) {
-            double var1 = (double)var0 / 256.0D;
-            double var3 = Math.asin(var1);
-            f[var0] = Math.cos(var3);
+            double var1 = (double) var0 / 256.0D;
+            double var3 = (fastMathMode == true ? (FastMath.asin(var1)) : (Math.asin(var1)));
+            f[var0] = (fastMathMode == true ? (FastMath.cosQuick(var3)) : (Math.cos(var3)));
```
Elierrr, line 311:
```suggestion
            f[var0] = Math.cos(var3);
```

The cosQuick method has even worse results than the cos method

NachoSpigot-Server/src/main/java/net/minecraft/server/MathHelper.java
```diff
@@ -286,7 +293,11 @@ public static double i(double var0) {
     static {
         int var0;
         for(var0 = 0; var0 < 65536; ++var0) {
-            b[var0] = (float)Math.sin((double)var0 * 3.141592653589793D * 2.0D / 65536.0D);
+            if (fastMathMode) {
+                b[var0] = (float) FastMath.sinQuick((double) var0 * FastMath.PI * 2.0D / 65536.0D);
+            } else {
+                b[var0] = (float) Math.sin((double) var0 * 3.141592653589793D * 2.0D / 65536.0D);
+            }
```
Elierrr, line 300:
```suggestion
           b[var0] = (float)Math.sin((double)var0 * Math.PI * 2.0D / 65536.0D);
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MathHelper.java
```diff
@@ -10,29 +12,32 @@
...
+    private static final boolean fastMathMode = NachoConfig.enableFastMath;
 
     public static float sin(float var0) {
-        return b[(int)(var0 * 10430.378F) & '\uffff'];
+        return (fastMathMode == true ? ((float) FastMath.sinQuick(b[(int)(var0 * 10430.378F) & '\uffff'])) : (b[(int)(var0 * 10430.378F) & '\uffff']));
     }
 
     public static float cos(float var0) {
-        return b[(int)(var0 * 10430.378F + 16384.0F) & '\uffff'];
+        return (fastMathMode == true ? ((float) FastMath.cosQuick(b[(int)(var0 * 10430.378F + 16384.0F) & '\uffff'])) : (b[(int)(var0 * 10430.378F + 16384.0F) & '\uffff']));
```
Elierrr, line 22:
```suggestion
        return b[(int)(var0 * 10430.378F + 16384.0F) & '\uffff'];
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MathHelper.java
```diff
@@ -10,29 +12,32 @@
     private static final double d;
     private static final double[] e;
     private static final double[] f;
+    private static final boolean fastMathMode = NachoConfig.enableFastMath;
 
     public static float sin(float var0) {
-        return b[(int)(var0 * 10430.378F) & '\uffff'];
+        return (fastMathMode == true ? ((float) FastMath.sinQuick(b[(int)(var0 * 10430.378F) & '\uffff'])) : (b[(int)(var0 * 10430.378F) & '\uffff']));
```
Elierrr, line 18:
```suggestion
        return b[(int)(var0 * 10430.378F) & '\uffff'];
```

Elierrr (changes requested):

Rastrian:
@Elierrr I need to know what kind of tests are you doing to understand why the change

Elierrr:
> @Elierrr I need to know what kind of tests are you doing to understand why the change

I'm comparing the results of the methods with multiple numbers.

Lucaskyy:
It seems that FastMath has worse precision, which causes calculations on the server-side to not be the same on the client-side. I believe MWHunter sent a video a while ago showing the issues it causes.

That's why it's not recommended to use FastMath for cos, sin and some others since these provide "wrong" values.

Lucaskyy:
Though, for what I know, these issues don't occur on other FastMath methods. Only cos and sin, hence @Elierrr suggested changes on removing those.

Rastrian:
> It seems that FastMath has worse precision, which causes calculations on the server-side to not be the same on the client-side. I believe MWHunter sent a video a while ago showing the issues it causes.
> 
> That's why it's not recommended to use FastMath for cos, sin and some others since these provide "wrong" values.

Yeah i have seen this, thats why this option is added as OPTIONAL, the sinQuick and cosQuick methods are ""faster"" than the sin and cos implementations.

The idea behind fastMath is not being totally precise, but fast.
Sample: https://gist.github.com/ijuma/840120

Lucaskyy:
How much faster is it _**now**_ though? Java has come a long way since Java 7, so the question is, is it still faster on Java 17?

Lucaskyy:
I have an idea, is FastMath used anywhere else **but** cos and sin?

If so, you could make an extra config for just the cos and sin parts since they can cause issues.

If not, then you can keep it like this imo.

Rastrian:
> How much faster is it _**now**_ though? Java has come a long way since Java 7, so the question is, is it still faster on Java 17? Since that gist benchmarks on Java 7, which is very old.

Im gonna make a test with the [jafama performance test](https://github.com/jasongilman/jafama-perf-test), and post in discord, but results keep the same since java 8.

Rastrian:
> I have an idea, is FastMath used anywhere else **but** cos and sin?
> 
> If so, you could make an extra config for just the cos and sin parts since they can cause issues.
> 
> If not, then you can keep it like this imo.

Yes, ceil, trigonometric methods, you can see some explanation under [changelog](https://github.com/jeffhain/jafama/blob/master/CHANGELOG.txt).

sadcenter:
> I think someone should test these changes in-game cause it's really massive PR

@Lucaskyy Have this in your mind

Rastrian:
> I have an idea, is FastMath used anywhere else **but** cos and sin?
> 
> If so, you could make an extra config for just the cos and sin parts since they can cause issues.
> 
> If not, then you can keep it like this imo.

added toggleable sin/cos.

NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
```diff
@@ -322,4 +322,46 @@ private static void disableFallAnimation() {
...
+    private static void enableFastMath() {
+        enableFastMath = getBoolean("settings.enable-fastmath", false);
+        c.addComment("settings.enable-fastmath", "Enable Fast Math usage, this sometimes break anticheats, be careful.");
+    }
+
+    public static boolean enableFastMathCosSin;
+
+    private static void enableFastMathCosSin() {
+        enableFastMathCosSin = getBoolean("settings.enable-fastmath-cos-sin", false);
+        c.addComment("settings.enable-fastmath-cos-sin", "Enable Fast Math usage with cos() and sin() methods.");
```
Elierrr, line 343:
```suggestion
        c.addComment("settings.enable-fastmath", "Enable Fast Math usage.");
    }

    public static boolean enableFastMathCosSin;

    private static void enableFastMathCosSin() {
        enableFastMathCosSin = getBoolean("settings.enable-fastmath-cos-sin", false);
        c.addComment("settings.enable-fastmath-cos-sin", "Enable Fast Math usage with cos() and sin() methods, this may break anticheats and server-side calculations.");

**Review by Elierrr — CHANGES_REQUESTED** [review]
Everything else LGTM

**Review by Elierrr — APPROVED** [review]

**Review by tomasalves8 — APPROVED** [review]
LGTM

**Comment by andreasdc** [comment]
@Rastrian maybe you would want to check this out: https://github.com/CobbleSword/NachoSpigot/issues/239

**Comment by andreasdc** [comment]
> > I have an idea, is FastMath used anywhere else **but** cos and sin?
> > If so, you could make an extra config for just the cos and sin parts since they can cause issues.
> > If not, then you can keep it like this imo.
> 
> Yes, ceil, trigonometric methods, you can see some explanation under [changelog](https://github.com/jeffhain/jafama/blob/master/CHANGELOG.txt).
> 
> When i added an PR here with fastmath for the knockback system, which one i was reworking, i said i don't recommend using FastMath for everything, its not an silver bullet.
> 
> Thats why I said, needs more work, but for now, if you want to use it, use it.
> 
> image
> 
> I'm gonna replace some methods before merge this PR.

What about difference in returns of methods?

File: NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java [hunk 22]
```diff
@@ -322,4 +322,46 @@ private static void disableFallAnimation() {
         c.addComment("settings.disabled-block-fall-animation", "Disables the fall animation for blocks");
     }
 
+    public static boolean disableInfinitSleeperThreadUsage;
+
+    private static void disableInfinitSleeperThreadUsage() {
+        disableInfinitSleeperThreadUsage = getBoolean("settings.disable-infinisleeper-thread-usage", false);
+        c.addComment("settings.disable-infinisleeper-thread-usage", "Disable infinisleeper thread usage, just enable this if you know what are you doing.");
```
Lucaskyy, line 329: Typo: `infinit`

NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
```diff
@@ -322,4 +322,46 @@ private static void disableFallAnimation() {
...
+    private static void titleEntityTickingTime() {
+        titleEntityTickingTime = getInt("settings.tile-entity-ticking-time", 20);
+        c.addComment("settings.tile-entity-ticking-time", "Ticking time in seconds for usage on tile entity operations.");
+    }
+
+    public static int itemDirtyTicks;
+
+    private static void itemDirtyTicks() {
+        itemDirtyTicks = getInt("settings.item-dirty-ticks", 20);
+        c.addComment("settings.item-dirty-ticks", "Change the behavior of dirty itemstacks checking.");
```
Lucaskyy, line 357: Is it possible to elaborate a bit more on what this does? (in the config comment)

NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
```diff
@@ -322,4 +322,46 @@ private static void disableFallAnimation() {
...
+
+    private static void enableFastMathCosSin() {
+        enableFastMathCosSin = getBoolean("settings.enable-fastmath-cos-sin", false);
+        c.addComment("settings.enable-fastmath-cos-sin", "Enable Fast Math usage with cos() and sin() methods, this may break anticheats and server-side calculations.");
+    }
+
+    public static int titleEntityTickingTime;
+
+    private static void titleEntityTickingTime() {
+        titleEntityTickingTime = getInt("settings.tile-entity-ticking-time", 20);
```
Lucaskyy, line 349: Is this right? Should it be 20 seconds according to the comment below?

Lucaskyy (changes requested):
Couple little things.

LewUwU:
I want to say somethings, explain every change because only saying in the title (fix(Class) optimize...) but how you optimized it? did you tested all the changes? all the ticking changes are dangerous, you tested properly if these values are correct?

NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
```diff
@@ -50,7 +50,7 @@ public static void init(File configFile) {
         File old_config = new File("nacho.json");
         if(old_config.exists()) migrate(old_config);
 
-        int configVersion = 2; // Update this every new configuration update
+        int configVersion = 3; // Update this every new configuration update
```
Elierrr, line 53:
```suggestion
        int configVersion = 4; // Update this every new configuration update
```
Config version updated in the #242

Elierrr (changes requested):

Elierrr (approved):

Rastrian:
> I want to say somethings, explain every change because only saying in the title (fix(Class) optimize...) but how you optimized it? did you tested all the changes? all the ticking changes are dangerous, you tested properly if these values are correct?

As far I tested, is working fine, thats why I asked some people to test also.
Test with redstone stress machines, stressing entity tiles, etc.

Rastrian:
> @Rastrian maybe you would want to check this out: #239

I think it should be on other PR, this already is massive.

Rastrian:
> > > I have an idea, is FastMath used anywhere else **but** cos and sin?
> > > If so, you could make an extra config for just the cos and sin parts since they can cause issues.
> > > If not, then you can keep it like this imo.
> > 
> > 
> > Yes, ceil, trigonometric methods, you can see some explanation under [changelog](https://github.com/jeffhain/jafama/blob/master/CHANGELOG.txt).
> > When i added an PR here with fastmath for the knockback system, which one i was reworking, i said i don't recommend using FastMath for everything, its not an silver bullet.
> > Thats why I said, needs more work, but for now, if you want to use it, use it.
> > image
> > I'm gonna replace some methods before merge this PR.
> 
> What about difference in returns of methods?

This difference it's subtle, can affect anticheat detection but doesn't completely affects the gameplay behavior.
But I think other people should test first instead of just accept it, some issues can be found with time, and thats why i said fastmath is not an silver bullet, but can be helpful in some situations.

NachoSpigot-Server/src/main/java/net/minecraft/server/Explosion.java
```diff
@@ -342,9 +344,9 @@ private void searchForBlocks(it.unimi.dsi.fastutil.longs.LongSet set, Chunk chun
             double stepZ = this.posZ;
 
             for (; f > 0.0F; f -= 0.22500001F) {
-                int floorX = org.bukkit.util.NumberConversions.floor(stepX);
-                int floorY = org.bukkit.util.NumberConversions.floor(stepY);
-                int floorZ = org.bukkit.util.NumberConversions.floor(stepZ);
+                int floorX = (NachoConfig.enableFastMath == true ? FastMath.floorToInt((Double.doubleToRawLongBits(stepX) >>> 63)) : org.bukkit.util.NumberConversions.floor(stepX));
```
Lucaskyy, line 347: Optional, but could you remove the ` == true` on all of them? Since they're useless since `enableFastMath` is a boolean anyway.

NachoSpigot-Server/src/main/java/net/minecraft/server/MathHelper.java
```diff
@@ -10,29 +12,33 @@
     private static final double d;
     private static final double[] e;
     private static final double[] f;
+    private static final boolean fastMathMode = NachoConfig.enableFastMath;
+    private static final boolean fastMathCosSin = NachoConfig.enableFastMathCosSin;
 
     public static float sin(float var0) {
-        return b[(int)(var0 * 10430.378F) & '\uffff'];
+        return (fastMathCosSin == true ? ((float) FastMath.sinQuick(b[(int)(var0 * 10430.378F) & '\uffff'])) : (b[(int)(var0 * 10430.378F) & '\uffff']));
```
Lucaskyy, line 19: Here too.

NachoSpigot-Server/src/main/java/net/minecraft/server/World.java
```diff
@@ -64,7 +68,7 @@ private void guard()
...
```
Lucaskyy, line 71: There is no need to qualify the entire class name, since you're importing it above.
```suggestion
    public final OptimizedWorldTileEntitySet tileEntityList = new OptimizedWorldTileEntitySet();
```

Lucaskyy (changes requested):
Looks good to me besides these little things. 👍 

Normally I'd do these myself, but since you disallowed others to modify this PR I can't do that so I have to let you do them.

Lucaskyy (approved):
LGTM!

Lucaskyy:
@Rastrian I would like to wait a little bit for hpfxd and/or Heath to review this too since it's a pretty big PR.
In the meantime, could you fix the conflicts? (probably config, easy to fix)

Lucaskyy:
@hpfxd @HeathLoganCampbell If any of you have the time, please review this PR in case I missed something.
If all is good, I'll merge it tomorrow.

Lucaskyy:
Also, just in case I haven't yet, I would like to thank @Rastrian very much for all he has done for this PR :)
Really appreciate the help!

hpfxd (dismissed):
I've not tested this in-game, but looks mostly good except for what I've pointed out.

Would prefer if your changes to upstream files had markers to indicate them, e.g. `// NachoSpigot start`, etc.

Lucaskyy (approved):

NachoSpigot-Server/src/main/java/me/rastrian/dev/OptimizedWorldTileEntitySet.java
```diff
@@ -0,0 +1,144 @@
...
+            }
+        }
+
+        return Iterators.concat(tileIterators.iterator());
+    }
+
+    private List<Class<? extends TileEntity>> getTileClassesToTick(long worldTime) {
+        List<Class<? extends TileEntity>> tilesToTick = new LinkedList<>();
+        for (Class<? extends TileEntity> registeredTileClass : this.registeredTiles.keySet()) {
+            long customTickInterval = OptimizedWorldTileEntitySet.CUSTOM_TICK_INTERVALS.get(registeredTileClass);
```
hpfxd, line 133: `get()` shoudn't be used, use the type-specific `getLong()` instead.
```suggestion
            long customTickInterval = OptimizedWorldTileEntitySet.CUSTOM_TICK_INTERVALS.getLong(registeredTileClass);
```

hpfxd (changes requested):
Server crashes upon startup, due to this exception.
```
java.lang.NullPointerException: Cannot invoke "java.lang.Long.longValue()" because the return value of "it.unimi.dsi.fastutil.objects.Object2LongMap.get(Object)" is null
        at me.rastrian.dev.OptimizedWorldTileEntitySet.getTileClassesToTick(OptimizedWorldTileEntitySet.java:133) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
        at me.rastrian.dev.OptimizedWorldTileEntitySet.tickIterator(OptimizedWorldTileEntitySet.java:120) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
        at net.minecraft.server.v1_8_R3.World.tickEntities(World.java:1729) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
        at net.minecraft.server.v1_8_R3.WorldServer.tickEntities(WorldServer.java:602) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
        at net.minecraft.server.v1_8_R3.MinecraftServer.B(MinecraftServer.java:863) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
        at net.minecraft.server.v1_8_R3.DedicatedServer.B(DedicatedServer.java:409) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
        at net.minecraft.server.v1_8_R3.MinecraftServer.A(MinecraftServer.java:724) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
        at net.minecraft.server.v1_8_R3.MinecraftServer.run(MinecraftServer.java:617) [NachoSpigot.jar:git-NachoSpigot-b15921f]
        at java.base/java.lang.Thread.run(Thread.java:833) [?:?]
```

This is caused by using `get()`, which returns `null` when an entry is not found as per the `java.util.Map` specification, which we try to cast to a primitive `long`.

Other than this, everything seems to be working as intended :)

Rastrian:
> Server crashes upon startup, due to this exception.
> 
> ```
> java.lang.NullPointerException: Cannot invoke "java.lang.Long.longValue()" because the return value of "it.unimi.dsi.fastutil.objects.Object2LongMap.get(Object)" is null
>         at me.rastrian.dev.OptimizedWorldTileEntitySet.getTileClassesToTick(OptimizedWorldTileEntitySet.java:133) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
>         at me.rastrian.dev.OptimizedWorldTileEntitySet.tickIterator(OptimizedWorldTileEntitySet.java:120) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
>         at net.minecraft.server.v1_8_R3.World.tickEntities(World.java:1729) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
>         at net.minecraft.server.v1_8_R3.WorldServer.tickEntities(WorldServer.java:602) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
>         at net.minecraft.server.v1_8_R3.MinecraftServer.B(MinecraftServer.java:863) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
>         at net.minecraft.server.v1_8_R3.DedicatedServer.B(DedicatedServer.java:409) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
>         at net.minecraft.server.v1_8_R3.MinecraftServer.A(MinecraftServer.java:724) ~[NachoSpigot.jar:git-NachoSpigot-b15921f]
>         at net.minecraft.server.v1_8_R3.MinecraftServer.run(MinecraftServer.java:617) [NachoSpigot.jar:git-NachoSpigot-b15921f]
>         at java.base/java.lang.Thread.run(Thread.java:833) [?:?]
> ```
> 
> This is caused by using `get()`, which returns `null` when an entry is not found as per the `java.util.Map` specification, which we try to cast to a primitive `long`.
> 
> Other than this, everything seems to be working as intended :)

I never had used fastutil collections, but thx.

hpfxd (approved):

CyberFlameGO (approved):
My goodness this is a big pull request. I'm glad to see that there's now a toggle for Jafama, and everything seems good to me

Lucaskyy:
Merging! Thank you very much for this PR :)

SaithTime:
FastMath cause a problem with projectiles

Lucaskyy:
@SaithTime In this PR?

SaithTime:
> @SaithTime Dans ce PR ?

Yes if this option "enable-fastmath-cos-sin" is on true

CyberFlameGO:
image

Lucaskyy:
> > @SaithTime Dans ce PR ?
> 
> Yes if this option "enable-fastmath-cos-sin" is on true

Config comments don't work yet, but the comment tells you not to use it unless you know what you're doing since it will break things. So don't use it.

Sneakometer:
I'm questioning the adding of FastMath in general. MathHelper already uses lookup tables. FastMath won't do any better and adding the check will probably cost more performance than gained. (Not really sure if the final static boolean is optimized away by the compiler at runtime though).

Am i missing something?

Rastrian:
> I'm questioning the addition of FastMath in general. MathHelper already uses lookup tables. FastMath won't do any better and adding the config check will probably end up costing more performance than gained. (Not really sure if the final static boolean is optimized away by the compiler at runtime though).
> 
> Am i missing something?

Yeah is optimized, its literally static, if an static boolean is concerning you about this change, maybe you should use another language instead of Java.

PR #234, opened 2021-11-13
Fixing twice explosion call.

### LINKED ISSUE #230: EntityExplodeEvent called twice
If a TNT explodes the EntityExplodeEvent is called twice first with a blockList containing only one AIR block and the second time with the actual blocklist.

```
    @EventHandler
    public void onEntityExplode(EntityExplodeEvent e) {
        System.out.println(e.blockList());
    }
```

Produces the following output

```
[01:56:01] [Server thread/INFO]: [CraftBlock{chunk=CraftChunk{x=-25z=8},x=-389,y=62,z=141,type=AIR,data=0}]
[01:56:01] [Server thread/INFO]: [CraftBlock{chunk=CraftChunk{x=-25z=8},x=-388,y=61,z=138,type=DIRT,data=0}, CraftBlock{chunk=CraftChunk{x=-25z=8},x=-390,y=65,z=143,type=DIRT,data=0}, ...}]
```
[nacho.txt](https://github.com/CobbleSword/NachoSpigot/files/7531148/nacho.txt)

sadcenter:
Fixes #230 by deleting useless event call (IMO).
We could also create a new event for this but I don't think it will be useful for anyone. 
Up for discussion


Lucaskyy:
I agree. Thank you for this PR! :)

PR #236, opened 2021-11-13
Recode of EnumProtocol & Suppliers

Lucaskyy:
Optimizing packet sending and receiving by using suppliers instead of Reflection.
Also remapped EnumProtocol to get a better understanding of it.

Since my recode broke ProtocolLib's packet registry, I've added a shim for it.
One caveat is that we now have an extra map, so extra memory has to be allocated.
There's a config option for this, so if you don't have ProtocolLib, I recommend you disable this.

### DIFF
--- a/NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
+++ b/NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
@@ -50,7 +50,7 @@ public class NachoConfig {
         File old_config = new File("nacho.json");
         if(old_config.exists()) migrate(old_config);
 
-        int configVersion = 2; // Update this every new configuration update
+        int configVersion = 3; // Update this every new configuration update
         version = getInt("config-version", configVersion);
         set("config-version", configVersion);
         c.setHeader(HEADER);
@@ -154,7 +154,7 @@ public class NachoConfig {
 
     private static <T> List getList(String path, T def) {
         config.addDefault(path, def);
-        return (List<T>) config.getList(path, config.getList(path));
+        return config.getList(path, config.getList(path));
     }
 
     private static String getString(String path, String def) {
@@ -320,6 +320,13 @@ public class NachoConfig {
     private static void disableFallAnimation() {
         disabledFallBlockAnimation = getBoolean("settings.disabled-block-fall-animation", false);
         c.addComment("settings.disabled-block-fall-animation", "Disables the fall animation for blocks");
+    }
+
+    public static boolean enableProtocolLibShim;
+
+    private static void enableProtocolLibShim() {
+        enableProtocolLibShim = getBoolean("settings.enable-protocollib-shim", true);
+        c.addComment("settings.enable-protocollib-shim", "Enable ProtocolLib network shim. Allows ProtocolLib to work, but requires extra memory. Disable this if you don't use ProtocolLib!");
     }
 
 }
--- a/NachoSpigot-Server/src/main/java/net/minecraft/server/ChunkProviderServer.java
+++ b/NachoSpigot-Server/src/main/java/net/minecraft/server/ChunkProviderServer.java
@@ -408,7 +408,7 @@ public class ChunkProviderServer implements IChunkProvider {
         // TacoSpigot start - use iterator for unloadQueue
         LongIterator iterator = unloadQueue.iterator();
         for (int i = 0; i < 100 && iterator.hasNext(); ++i) {
-            long chunkcoordinates = iterator.next();
+            long chunkcoordinates = iterator.nextLong();
             iterator.remove();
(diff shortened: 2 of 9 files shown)


andreasdc:
Fck Plib 🚀

LewUwU:
> WIP, don't use. ProtocolLib not working at the moment, needs fixing.

Fixed in my pr, needs merge

Lucaskyy:
> > WIP, don't use. ProtocolLib not working at the moment, needs fixing.
> 
> Fixed in my pr, needs merge

Hmm? I can't see a PR anywhere.

LewUwU:
> > > WIP, don't use. ProtocolLib not working at the moment, needs fixing.
> > 
> > 
> > Fixed in my pr, needs merge
> 
> Hmm? I can't see a PR anywhere.

https://github.com/dmulloy2/ProtocolLib/pulls

Lucaskyy:
Oh, I see. So the issue is Netty, not my code xP

Lucaskyy:
Not mergeable yet, currently debugging an issue causing the EnumProtocol attribute to be null.

Lucaskyy:
Some little benchmarks. My PC is not that good (Intel i5) and I have a shit ton of programs open, so these can make the benchmarks worse.

Before:
Before Img

After:
After Img

Lucaskyy:
ProtocolLib is still not working, it seems that I'll have to create a shim for it.

Lucaskyy:
This PR is finished. If people have the time, please review this PR so I can fix any mistakes I might've made. :)

Elierrr (approved):
LGTM!

hpfxd:
Here are my results from testing this.
- 150 bots, jumping around a superflat world in a 64x64 area.
- Mostly default configuration, except for encryption being disabled.
- aarch64 machine, 4GB memory allocated
  - I also tested this on my personal machine (Ryzen 5 5600x, 20 GB memory allocated), but profiling accurately on Windows is challenging, although results were consistent with the shown results.

**Entity tracking**
| Baseline | This PR |
| -------- | ------- |
| img | img |

**Netty event group**
| Baseline | This PR |
| -------- | ------- |
| img | img |

As seen, this PR seems to improve performance slightly during packet sending.

I have concerns about plugin compatibility, due to field renaming and removal.  
I see you've already merged this, but please test it with the following plugins & libraries if you have time:
- [ViaVersion](https://github.com/ViaVersion/ViaVersion)
- [TinyProtocol](https://github.com/dmulloy2/ProtocolLib/blob/master/TinyProtocol/src/main/java/com/comphenix/tinyprotocol/TinyProtocol.java)
- [packetevents](https://github.com/retrooper/packetevents)

Lucaskyy:
@hpfxd I've tested it with ViaVersion, no issues.
ProtocolLib had issues (due to not using BiMap anymore), so I created a shim for it (read-only).
I'll test the other plugins too.

Lucaskyy:
As for testing packetevents, I'll also put Vulcan on it to see it in action (better testing)

Lucaskyy:
@hpfxd Sorry, took a bit longer than expected.
Just installing PacketEvents works, don't get any errors or whatever.
ProtocolLib works just fine too, and so does ViaVersion.

andreasdc:
image

Lucaskyy:
@andreasdc I'm not sure if something said bye-bye on your computer, but it definitely shouldn't be slower than Reflection.

andreasdc:
> @andreasdc I'm not sure if something said bye-bye on your computer, but it definitely shouldn't be slower than Reflection.

What do you mean by that? It's profiler from production.

Lucaskyy:
Also, @hpfxd I tested using ThotPatrol (another AC) which uses ProtocolLib instead of PacketEvents, and that worked fine.
So I think this PR should work just fine, but if any issues occur I'll fix them.

Lucaskyy:
> > @andreasdc I'm not sure if something said bye-bye on your computer, but it definitely shouldn't be slower than Reflection.
> 
> What do you mean by that? It's profiler from production.

I don't trust that profiler report at all, since there's no way that suppliers would be slower than Reflection. :P

andreasdc:
> > > @andreasdc I'm not sure if something said bye-bye on your computer, but it definitely shouldn't be slower than Reflection.
> > 
> > 
> > What do you mean by that? It's profiler from production.
> 
> I don't trust that profiler report at all, since there's no way that suppliers would be slower than Reflection. :P
> 
> Or, in other words, I'm very confused that this would be slower.
> 
> Could you instead do the same as hpfxd, and try the latest production build and a couple of builds earlier before this was added? Since you can't guarantee that you will get the same results on another day.

You had earlier profiler and current, I will make sure it has the newest changes tomorrow, maybe there was a little more players too.

andreasdc:
@Lucaskyy 
image

PR #252, opened 2021-11-17
Remove FastMath sin and cos config

### LINKED ISSUE #244: Fastmath breaks projectiles
Setting `enable-fastmath-cos-sin` to `true` causes projectiles such as bows, eggs and enderpearls to fly into wrong directions when shooting them in north or east direction.

Elierrr:
# Description

This PR removes FastMath's con and sin methods. This is done due to the fact that FastMath's con and sin gives [less performance and inaccurate results](https://github.com/CobbleSword/NachoSpigot/files/7550597/fastmathvsvanilla.txt). 

Fixes #244 

# How has this been tested?

I have tested #244.

# Checklist:

- [x] I have tested my code.

### DIFF
--- a/NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
+++ b/NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
@@ -50,7 +50,7 @@ public class NachoConfig {
         File old_config = new File("nacho.json");
         if(old_config.exists()) migrate(old_config);
 
-        int configVersion = 5; // Update this every new configuration update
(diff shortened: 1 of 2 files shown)


refractored:
perhaps add a comment that says it degrades performance instead?

Elierrr:
> perhaps add a comment that says it degrades performance instead?

There's no point in even having it if all it does is cause issues and decrease performance.

refractored:
makes sense

Lucaskyy:
I agree with this. Until this shows that it's worth keeping, I'll remove it.
Thanks Elierrr!

PR #254, opened 2021-11-17
Async Knockback and Hit Detection

Rastrian:
# Description

Added structures to enable async knockback and hit detection modifications.
This basically is used by most of bigger PvP servers, pratically just change the behavior of how hits and knockback packets will perform in network manager.

- [X] Async Knockback.
- [X] Async Hit Detection.
- [X] Adjustable TPS of Knockback and Hit Detection threads.
- [X] Support to /tps.

# How has this been tested?

Tested PVP In-game, you can notice some improvement above 60-80 ms.
Help pvp servers with high player count pvps also, with a good performance above other packets.

If you don't understand how TPS relation works in minecraft, i recommend you watch this video:
[Tickrate proof](https://www.youtube.com/watch?v=iKri-UEwHk8 "Tickrate proof")

You can see also an spigot discussion about this: https://www.spigotmc.org/threads/over-20-tps.95756/ (i have commented in this 6y ago kekw)

# Additional Comments

This project is a draft, and need some testing, if you wan't to contribute with some code review, it will help me a lot.
You can get an pre-compiled jar with changes [clicking here, to access our github actions](https://github.com/Argarian-Network/NachoSpigot/actions?query=branch%3Aasync-kb-hit).

# Checklist:

- [x] I have reviewed my code thoroughly.
- [x] I have tested my code.
- [x] My changes generate no new warnings

PR #248, opened 2021-11-16
Async Entity Tracker

Rastrian:
Implemented Async Entity Tracker, + 2 config options (disable entity tracker and adjust threads).

You can test by yourself or getting an compiled jar into actions: https://github.com/Argarian-Network/NachoSpigot/actions?query=branch%3Aasync-entity-tracker

PR #256, opened 2021-11-17
fix comments and enforce proper usage of them

Lucaskyy:
# Description

Comments now show up in the config and a new way of comments is used and enforced during tests.
If the old way of comments is still used, tests will fail and it will tell you what is wrong.
They must now be specified in the `loadComments` method.

Thanks, @Elierrr for fixing it!

# How has this been tested?

Tested locally, comments work.
Enforcing also works, tests fail if the old method is used anywhere.

# Checklist:

- [x] I have tested my code.


Lucaskyy:
#248 #254 Please merge master when this is merged into your PR and update the comment locations, or tests will fail!

PR #263, opened 2021-11-24
Don't re-run ViaVersion's patch

Elierrr:
# Description

Removes the ViaVersion block patch.

# Additional Comments

This patch was added since ViaVersion wouldn't run this patch on NachoSpigot. However, in https://github.com/ViaVersion/ViaVersion/pull/2561, this issue is resolved and this patch is no longer needed.

# Checklist:

- [x] I have tested my code.

### DIFF
--- a/NachoSpigot-Server/src/main/java/dev/cobblesword/nachospigot/Nacho.java
+++ b/NachoSpigot-Server/src/main/java/dev/cobblesword/nachospigot/Nacho.java
@@ -66,10 +66,4 @@ public class Nacho {
     }
 
     public Set<MovementListener> getMovementListeners() { return movementListeners; }
-
-    public void applyPatches() {
-        // Nacho start - [Nacho-0041] Fix block placement
-        RuntimePatches.applyViaVersionBlockPatch();
-        // Nacho end
-    }
 }
--- a/NachoSpigot-Server/src/main/java/net/minecraft/server/DedicatedServer.java
+++ b/NachoSpigot-Server/src/main/java/net/minecraft/server/DedicatedServer.java
@@ -340,8 +340,6 @@ public class DedicatedServer extends MinecraftServer implements IMinecraftServer
                     thread1.start();
                 } */
 
-                Nacho.get().applyPatches(); // Nacho
-
                 return true;
             }
         }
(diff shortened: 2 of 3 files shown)


CyberFlameGO:
Shouldn't this just be turned into a config option, for people on older Citizens2 builds?

Elierrr:
> Shouldn't this just be turned into a config option, for people on older ViaVersion builds?

There would be no reason to be on older builds since newer ones would support newer MC versions

CyberFlameGO:
I'm talking in the sense of forks, and newer builds may not be relevant to some people because 1.8 is pretty much obsolete (sadly)

Elierrr:
> I'm talking in the sense of forks, and newer builds may not be relevant to some people because 1.8 is pretty much obsolete (sadly)

What do you mean by the sense of forks?

xIsm4:
> There would be no reason to be on older builds since newer ones would support newer MC versions

That's no way true, Spigot 1.7/1.8x still used a lot.
Also who will use an outdated version.

Lucaskyy:
@xIsm4 I think you understood his comment wrong.
I think he means that newer builds of ViaVersion support the newest MC versions.
For example, 2.0 might not support 1.17, but 2.1 does so you'd want 2.1 most of the time to support the latest versions.

Lucaskyy:
@Elierrr I worded this wrong, it isn't a patch at all.

All it actually does is fix block placement when using ViaVersion and you don't have TacoSpigot as brand name.
This happens on ALL versions of ViaVersion, The reason is, ViaVersion only activates their block placement fix when TacoSpigot is in the MC brand, and since this is NachoSpigot, that is not the case. So we've got to activate it ourselves, manually.
That's what this fix does, so we should keep it.

Thanks anyway, though!

Elierrr:
> The reason is, ViaVersion only activates their block placement fix when TacoSpigot is in the MC brand, and since this is NachoSpigot, that is not the case.

In https://github.com/ViaVersion/ViaVersion/pull/2561, instead of checking if it's Paper 1.8.8 using server brand, it uses Paper's config class meaning that NachoSpigot is detected as Paper 1.8.8 and the ViaVersion's fix is enabled which is why I made this PR. @Lucaskyy

Lucaskyy:
Oh, my bad, I didn't read the PR message fully. Sorry!

Lucaskyy:
I'll merge this then, thank you for this PR! :)

PR #354, opened 2022-02-20
Clear player lag compensator cache on disconnect

notdap:
# Description

Fixes an issue where a player's lag compensator cache would not be cleared after they disconnected.

# How has this been tested?

I haven't tested it but it's simple enough. 

# Checklist:

- [ ] I have tested my code.


Lucaskyy:
Oh, I didn't notice that myself. That's a pretty bad memory leak I think. Anyway, thank you for this PR! :)

PR #359, opened 2022-02-25
add disableDisconnectSpam config

crafter23456:
# Description

Adds the option to disable that players can get kicked because of disconnect.spam. Mostly because of higher clients. The option doesnt include the line 1217 of exactly that file because I guess thats the real spam preventer and the other code which kicks for disconnect.spam in line 2132 which can be deactivated, is the detector of the chat how fast a user can type.

# How has this been tested?

i tested it whith typing and check if it triggers or not

# Checklist:

- [x] I have tested my code.

### DIFF
--- a/NachoSpigot-Server/src/main/java/net/minecraft/server/PlayerConnection.java
+++ b/NachoSpigot-Server/src/main/java/net/minecraft/server/PlayerConnection.java
@@ -2127,25 +2127,27 @@ public class PlayerConnection implements PacketListenerPlayIn, IUpdatePlayerList
         }
         // CraftBukkit end
     }
-
+	
+	
     public void a(PacketPlayInTabComplete packetplayintabcomplete) {
-        PlayerConnectionUtils.ensureMainThread(packetplayintabcomplete, this, this.player.u());
-        // CraftBukkit start
-        if (chatSpamField.addAndGet(this, 10) > 500 && !this.minecraftServer.getPlayerList().isOp(this.player.getProfile())) {
-            this.disconnect("disconnect.spam");
-            return;
+        if (NachoConfig.disableDisconnectSpam == false) {
+            PlayerConnectionUtils.ensureMainThread(packetplayintabcomplete, this, this.player.u());
+            // CraftBukkit start
+            if (chatSpamField.addAndGet(this, 10) > 500 && !this.minecraftServer.getPlayerList().isOp(this.player.getProfile())) {
+                this.disconnect("disconnect.spam");
+                return;
+            }
+            // CraftBukkit end
+            ArrayList arraylist = Lists.newArrayList();
+            Iterator iterator = this.minecraftServer.tabCompleteCommand(this.player, packetplayintabcomplete.a(), packetplayintabcomplete.b()).iterator();
+
+            while (iterator.hasNext()) {
+                String s = (String) iterator.next();
+                arraylist.add(s);
+            }
+
+            this.player.playerConnection.sendPacket(new PacketPlayOutTabComplete((String[]) arraylist.toArray(new String[arraylist.size()])));
         }
-        // CraftBukkit end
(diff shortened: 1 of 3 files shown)


Lucaskyy:
This implementation is flawed. Enabling the option would disable tab-completion looking at the diff, which is not what we're trying to achieve here.

crafter23456:
I wanted to archive that. That tab check is for 1.13+ player a mess. even if I'm in the chat and hold my finger on a button, after a bit I get kicked. that PR prevents it if its activated. normally it doesnt change anything. the normal disconnect.spam with spamming the chat should be keept so I only did the tab thing which causes it.

LewUwU:
https://github.com/CobbleSword/NachoSpigot/pull/359/commits/80c426e8069f0bff0286f708153251545098c163#diff-6ef3cc74793f80be4e7a7987b130f887b43c901c7e01d785ecab39de3c0587aeR2136

Also use ! instead of using == false

crafter23456:
> [80c426e#diff-6ef3cc74793f80be4e7a7987b130f887b43c901c7e01d785ecab39de3c0587aeR2136](https://github.com/CobbleSword/NachoSpigot/commit/80c426e8069f0bff0286f708153251545098c163#diff-6ef3cc74793f80be4e7a7987b130f887b43c901c7e01d785ecab39de3c0587aeR2136)
> 
> Also use ! instead of using == false

i changed it to !, is there a performance reason behind it or only better to look?

Lucaskyy:
There's no performance reason behind it, just a convention.

Lucaskyy:
> That tab check is for 1.13+ player a mess.

Then shouldn't we fix it instead of disabling it for everyone?

crafter23456:
> Then shouldn't we fix it instead of disabling it for everyone?

I mean it shouldnt be a loss that player get kicked falsely because of typing too fast (not sending!). the other reason for disconnect.spam is still present, if player spam (with sending!). That should be.

LewUwU:
Your check should be done in the chatSpamField verification, not in the whole handler

NachoSpigot-Server/src/main/java/net/minecraft/server/PlayerConnection.java
```diff
@@ -2127,21 +2127,23 @@ public void a(PacketPlayInAbilities packetplayinabilities) {
...
-            return;
+        // CraftBukkit start // Nacho start
+        if (!NachoConfig.disableDisconnectSpam) {
+            if (chatSpamField.addAndGet(this, 10) > 500 && !this.minecraftServer.getPlayerList().isOp(this.player.getProfile())) {
+                this.disconnect("disconnect.spam");
+                return;
+            }
         }
-        // CraftBukkit end
+        // Nacho end // CraftBukkit end
```
Elierrr, line 2141:
```suggestion
        // CraftBukkit/Nacho end
```

NachoSpigot-Server/src/main/java/net/minecraft/server/PlayerConnection.java
```diff
@@ -2127,21 +2127,23 @@ public void a(PacketPlayInAbilities packetplayinabilities) {
...
-
+	
+	
     public void a(PacketPlayInTabComplete packetplayintabcomplete) {
         PlayerConnectionUtils.ensureMainThread(packetplayintabcomplete, this, this.player.u());
-        // CraftBukkit start
-        if (chatSpamField.addAndGet(this, 10) > 500 && !this.minecraftServer.getPlayerList().isOp(this.player.getProfile())) {
-            this.disconnect("disconnect.spam");
-            return;
+        // CraftBukkit start // Nacho start
```
Elierrr, line 2134:
```suggestion
        // CraftBukkit/Nacho start
```

Elierrr (approved):

PR #371, opened 2022-03-04
Fix for player#setFlying not changing the player's flying status

termanator1128:
# Description

I have removed the check from setFlying to update the player's flying state. The current way prevents setting the player's flying state. IE: player.setFlying(true) would not make the player fly.

# How has this been tested?

I just made a simple plugin when joining the server to set setAllowFlight and setFlying to true.

# Checklist:

- [x] I have tested my code.


LewUwU (changes requested):
Not a proper fix

sadcenter:
Can fly = Can a player double click the space to start flying?
Is flying = Is player flying in the air?

MWHunter:
It is valid protocol for the server to set:
- Player can fly false
- Player is flying true

It locks the player into the state of flying.  However, this isn't intuitive.

PR #372, opened 2022-03-05
Thread Affinity

windcolor-dev:
# Description

This adds configurable thread affinity, it increases performance by giving the server a whole cpu core to run on, but other processes cannot use this core. (See https://github.com/OpenHFT/Java-Thread-Affinity)

# How has this been tested?

I have not had a chance to test this yet.

# Checklist:

- [ x ] I have reviewed my code thoroughly.
- [ x ] My changes generate no new warnings
- [    ] I have tested my code.

### DIFF
--- a/NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
+++ b/NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
@@ -50,7 +50,7 @@ public class NachoConfig {
         File old_config = new File("nacho.json");
         if(old_config.exists()) migrate(old_config);
 
-        int configVersion = 6; // Update this every new configuration update
+        int configVersion = 7; // Update this every new configuration update
         version = getInt("config-version", configVersion);
         set("config-version", configVersion);
         c.setHeader(HEADER);
--- a/NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
+++ b/NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
@@ -143,6 +143,7 @@ public class NachoConfig {
         c.addComment("settings.commands.enable-help-command", "Toggles the /help command");
         c.addComment("settings.use-improved-hitreg", "Enables the usage of an improved hitreg based on lag compensation and small other details.");
         c.addComment("settings.disable-disconnect-spam", "Disables that players can be kicked because of disconnect.spam.");
+        c.addComment("settings.thread-affinity", "Allocates a cpu to the server, increases performance but uses a whole cpu.");
         NachoWorldConfig.loadComments();
     }
 
@@ -403,5 +404,11 @@ public class NachoConfig {
 
     private static void disableDisconnectSpam() {
         disableDisconnectSpam = getBoolean("settings.disable-disconnect-spam", false);
+    }
+    
+    public static boolean threadAffinity;
+
+    private static void threadAffinity() {
+        threadAffinity = getBoolean("settings.thread-affinity", false);
     }
 }
--- a/NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
+++ b/NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
@@ -143,7 +143,7 @@ public class NachoConfig {
         c.addComment("settings.commands.enable-help-command", "Toggles the /help command");
         c.addComment("settings.use-improved-hitreg", "Enables the usage of an improved hitreg based on lag compensation and small other details.");
         c.addComment("settings.disable-disconnect-spam", "Disables that players can be kicked because of disconnect.spam.");
-        c.addComment("settings.thread-affinity", "Allocates a cpu to the server, increases performance but uses a whole cpu.");
+        c.addComment("settings.thread-affinity", "Allocates a cpu core to the server, increases performance but uses a whole core.");
         NachoWorldConfig.loadComments();
     }
 
--- a/NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
+++ b/NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
@@ -576,27 +576,22 @@ public abstract class MinecraftServer implements Runnable, ICommandListener, IAs
                 //long lastTick = System.nanoTime(), catchupTime = 0, curTime, wait, tickSection = lastTick;
                 long start = System.nanoTime(), lastTick = start - TICK_TIME, catchupTime = 0, curTime, wait, tickSection = start;
                 // PaperSpigot end
-                
-             // Enable thread affinity
-        		if (NachoConfig.threadAffinity)
-        		{
-        			System.out.println(" ");
-        			System.out.println("Enabling Thread Affinity...");
-        			lock = AffinityLock.acquireLock();
-        			if (lock.cpuId() != -1)
-        			{
-        				System.out.println("CPU " + lock.cpuId() + " locked for server usage.");
-        				System.out.println("This will boost the server's performance, but will use more cpu.");
-        				System.out.println("This is most effective on linux with JNA installed.");
-        				System.out.println("See https://github.com/OpenHFT/Java-Thread-Affinity");
-        				System.out.println(" ");
-        			} else
-        			{
-        				System.out.println("An error occured whilst enabling thread affinity!");
-        				System.out.println(" ");
-        			}
-
-        		}
+                // Enable thread affinity
+                if (NachoConfig.threadAffinity) {
+                    MinecraftServer.LOGGER.info(" ");
+                    MinecraftServer.LOGGER.info("Enabling Thread Affinity...");
+                    lock = AffinityLock.acquireLock();
+                    if (lock.cpuId() != -1) {
+                        MinecraftServer.LOGGER.info("CPU " + lock.cpuId() + " locked for server usage.");
+                        MinecraftServer.LOGGER.info("This will boost the server's performance, but will use more cpu.");
+                        MinecraftServer.LOGGER.info("This is most effective on linux with JNA installed.");
+                        MinecraftServer.LOGGER.info("See https://github.com/OpenHFT/Java-Thread-Affinity");
+                        MinecraftServer.LOGGER.info(" ");
+                    } else {
+                        MinecraftServer.LOGGER.error("An error occured whilst enabling thread affinity!");
+                        MinecraftServer.LOGGER.error(" ");
+                    }
+                }
                 while (this.isRunning) {
                     curTime = System.nanoTime();
                     // PaperSpigot start - Further improve tick loop
--- a/NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
+++ b/NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
@@ -557,10 +557,9 @@ public abstract class MinecraftServer implements Runnable, ICommandListener, IAs
     }
     // PaperSpigot End
  
-    // Thread affinity - lock
-    AffinityLock lock = null;
     
     public void run() {
+        AffinityLock lock = null; // Nacho
         try {
             if (this.init()) {
                 this.ab = az();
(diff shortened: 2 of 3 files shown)


NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
```diff
@@ -143,6 +143,7 @@ static void loadComments() {
         c.addComment("settings.commands.enable-help-command", "Toggles the /help command");
         c.addComment("settings.use-improved-hitreg", "Enables the usage of an improved hitreg based on lag compensation and small other details.");
         c.addComment("settings.disable-disconnect-spam", "Disables that players can be kicked because of disconnect.spam.");
+        c.addComment("settings.thread-affinity", "Allocates a cpu to the server, increases performance but uses a whole cpu.");
```
crafter23456, line 146:
```suggestion
        c.addComment("settings.thread-affinity", "Allocates a cpu core to the server, increases performance but uses a whole core.");
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
```diff
@@ -571,6 +576,27 @@ public void run() {
...
+        				System.out.println("This is most effective on linux with JNA installed.");
+        				System.out.println("See https://github.com/OpenHFT/Java-Thread-Affinity");
+        				System.out.println(" ");
+        			} else
+        			{
+        				System.out.println("An error occured whilst enabling thread affinity!");
+        				System.out.println(" ");
+        			}
+
+        		}
```
crafter23456, line 599:
```suggestion
                // Enable thread affinity
                if (NachoConfig.threadAffinity) {
                    System.out.println(" ");
                    System.out.println("Enabling Thread Affinity...");
                    lock = AffinityLock.acquireLock();
                    if (lock.cpuId() != -1) {
                        System.out.println("CPU " + lock.cpuId() + " locked for server usage.");
                        System.out.println("This will boost the server's performance, but will use more cpu.");
                        System.out.println("This is most effective on linux with JNA installed.");
                        System.out.println("See https://github.com/OpenHFT/Java-Thread-Affinity");
                        System.out.println(" ");
                    } else {
                        System.out.println("An error occured whilst enabling thread affinity!");
                        System.out.println(" ");
                    }
                }
```
better formatting in relation to the other code and i replaced the tab-stops

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
```diff
@@ -651,6 +677,11 @@ public void run() {
 
             this.a(crashreport);
         } finally {
+        	// Release thread affinity lock
+        	if (lock != null)
+        	{
+        		lock.release();
+        	}
```
crafter23456, line 684:
```suggestion
            // Release thread affinity lock
            if (lock != null)  {
                lock.release();
            }
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
```diff
@@ -555,6 +557,9 @@ public double getAverage() {
     }
     // PaperSpigot End
  
+	// Thread affinity - lock
```
crafter23456, line 560:
```suggestion
    // Thread affinity - lock
```

LewUwU (changes requested):
Use the MinecraftServer Logger instead of using System.out.println

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
windcolor-dev: Yeah sorry that is how I'm used to doing my brackets. I believe that the info on JNA should stay there but ig some of the other stuff can be shaved off.

crafter23456: yeah its also a mess for me :D tab seems fine local but GH dont likes them and format it different

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
```diff
@@ -571,6 +576,22 @@ public void run() {
                 //long lastTick = System.nanoTime(), catchupTime = 0, curTime, wait, tickSection = lastTick;
                 long start = System.nanoTime(), lastTick = start - TICK_TIME, catchupTime = 0, curTime, wait, tickSection = start;
                 // PaperSpigot end
+                // Enable thread affinity
+                if (NachoConfig.threadAffinity) {
+                    MinecraftServer.LOGGER.info(" ");
+                    MinecraftServer.LOGGER.info("Enabling Thread Affinity...");
```
Elierrr, line 582:
```suggestion
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
```diff
@@ -571,6 +576,22 @@ public void run() {
...
+                    lock = AffinityLock.acquireLock();
+                    if (lock.cpuId() != -1) {
+                        MinecraftServer.LOGGER.info("CPU " + lock.cpuId() + " locked for server usage.");
+                        MinecraftServer.LOGGER.info("This will boost the server's performance, but will use more cpu.");
+                        MinecraftServer.LOGGER.info("This is most effective on linux with JNA installed.");
+                        MinecraftServer.LOGGER.info("See https://github.com/OpenHFT/Java-Thread-Affinity");
+                        MinecraftServer.LOGGER.info(" ");
```
Elierrr, line 589:
```suggestion
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
windcolor-dev: Nice to know :)

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
```diff
@@ -571,6 +576,22 @@ public void run() {
...
+                    } else {
+                        MinecraftServer.LOGGER.error("An error occured whilst enabling thread affinity!");
+                        MinecraftServer.LOGGER.error(" ");
```
Elierrr, line 592:
```suggestion
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
```diff
@@ -571,6 +576,22 @@ public void run() {
                 //long lastTick = System.nanoTime(), catchupTime = 0, curTime, wait, tickSection = lastTick;
                 long start = System.nanoTime(), lastTick = start - TICK_TIME, catchupTime = 0, curTime, wait, tickSection = start;
                 // PaperSpigot end
+                // Enable thread affinity
+                if (NachoConfig.threadAffinity) {
+                    MinecraftServer.LOGGER.info(" ");
+                    MinecraftServer.LOGGER.info("Enabling Thread Affinity...");
+                    lock = AffinityLock.acquireLock();
+                    if (lock.cpuId() != -1) {
+                        MinecraftServer.LOGGER.info("CPU " + lock.cpuId() + " locked for server usage.");
```
Elierrr, line 585:
```suggestion
                        MinecraftServer.LOGGER.info("Nacho: CPU " + lock.cpuId() + " locked for server usage.");
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
```diff
@@ -555,6 +557,9 @@ public double getAverage() {
     }
     // PaperSpigot End
  
+    // Thread affinity - lock
+    AffinityLock lock = null;
+    
     public void run() {
```
Elierrr, line 563:
```suggestion
    
    public void run() {
        AffinityLock lock = null; // Nacho
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
```diff
@@ -571,6 +576,22 @@ public void run() {
                 //long lastTick = System.nanoTime(), catchupTime = 0, curTime, wait, tickSection = lastTick;
                 long start = System.nanoTime(), lastTick = start - TICK_TIME, catchupTime = 0, curTime, wait, tickSection = start;
                 // PaperSpigot end
+                // Enable thread affinity
```
Elierrr, line 579:
```suggestion
                // Nacho start
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
```diff
@@ -571,6 +576,22 @@ public void run() {
...
+                        MinecraftServer.LOGGER.info("CPU " + lock.cpuId() + " locked for server usage.");
+                        MinecraftServer.LOGGER.info("This will boost the server's performance, but will use more cpu.");
+                        MinecraftServer.LOGGER.info("This is most effective on linux with JNA installed.");
+                        MinecraftServer.LOGGER.info("See https://github.com/OpenHFT/Java-Thread-Affinity");
+                        MinecraftServer.LOGGER.info(" ");
+                    } else {
+                        MinecraftServer.LOGGER.error("An error occured whilst enabling thread affinity!");
+                        MinecraftServer.LOGGER.error(" ");
+                    }
+                }
```
Elierrr, line 594:
```suggestion
                }
                // Nacho end
```

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
```diff
@@ -651,6 +672,10 @@ public void run() {
 
             this.a(crashreport);
         } finally {
+            // Release thread affinity lock
+            if (lock != null)  {
+                lock.release();
+            }
```
Elierrr, line 678:
```suggestion
            // Nacho start
            if (lock != null)  {
                lock.release();
            }
            // Nacho end
```

Elierrr (changes requested):

CyberFlameGO:
I agree with Elier’s suggested edits - we don’t need to pollute the console

sadcenter:
Please, benchmark and test this before merging.

NachoSpigot-Server/src/main/java/net/minecraft/server/MinecraftServer.java
LewUwU: This should be keep out of the method since MinecraftServer itself is a runnable so you're creating a AffinityLock instance for every run by doing this

windcolor-dev:
I did some **very very basic** benchmarking by spawning 100 pigs via a plugin. This was tested on Linux with JNA installed. 

Here are the results.

With thread affinity: https://spark.lucko.me/A2Ti3hEt6v
Without thread affinity: https://spark.lucko.me/xLyLqEpHdZ

If you know of a better way to benchmark please let me know :)

MWHunter:
I'd suggest benchmarking using chunky over many hours.

HeathLoganCampbell:
mmh What's the bench mark of this because I'm pretty sure it isn't that large?

Lucaskyy:
I really like the idea of this PR. Should at least make the server run smoother since there's no thread switching on the main thread when heavy multi-threaded things are done.

windcolor-dev:
> I'd suggest benchmarking using chunky over many hours.

Good idea :)

windcolor-dev:
> mmh What's the benchmark, what area does it try and target?

I was trying to put some stress on the server via entities, with and without thread affinity to determine the server performance (MSPT, TPS)

windcolor-dev:
> I really like the idea of this PR. Should at least make the server run smoother since there's no thread switching on the main thread when heavy multi-threaded things are done.

Thank you :)

caoli5288:
The Thread Affinity is usually used to reduce the delay of interrupt or schedule **AND DAMAGES THROUGHPUT**. It's hard to say that MC servers need such features.

caoli5288:
It's need user modify the grub commandline. In fact, this is very impractical for common user.

> Java-Thread-Affinity requires that you first isolate some CPU’s.
> Once a CPU core is isolated, the Linux scheduler will not use the CPU core to run any user-space processes. The isolated CPUs will not participate in load balancing, and will not have tasks running on them unless explicitly assigned.
> To isolate the 1st and 3rd CPU cores (CPU numbers start from 0) on your system, add the following to the kernel command line during boot:
>isolcpus=1,3

In fact, if you don't do this step, it will only do harm and no good for performance: your thread can't get free CPU, and the OS will take away your CPU time.

MWHunter:
From the bloom discord: https://cdn.discordapp.com/attachments/741770743272046612/815727694716731402/Screenshot_20210228-124613_Sheets.jpg

Thread pinning improves consistency when done right, but it does hurt throughput.

windcolor-dev:
> It's need user modify the grub commandline. In fact, this is very impractical for common user.
> 
> > Java-Thread-Affinity requires that you first isolate some CPU’s.
> > Once a CPU core is isolated, the Linux scheduler will not use the CPU core to run any user-space processes. The isolated CPUs will not participate in load balancing, and will not have tasks running on them unless explicitly assigned.
> > To isolate the 1st and 3rd CPU cores (CPU numbers start from 0) on your system, add the following to the kernel command line during boot:
> > isolcpus=1,3
> 
> In fact, if you don't do this step, it will only do harm and no good for performance: your thread can't get free CPU, and the OS will take away your CPU time.

Interesting, I guess this makes this feature irrelevant for average users

PR #381, opened 2022-03-21
New ticker and NetworkManager was updated

Dysaido:
# Description

Upgrade tick counter from MinecraftServer, upgrade queue from NetworkManager and Scheduler from bukkit. It is ReentrantBlockableEventLoop from 1.16.5 paper and it works better then 1.8.8 counter. I updated timings too. It works as good as before, just the protocolsupport can't running this version.

# How has this been tested?

I tested it on my local with 2 account and I used fawe, I think the protocolsupport doesn't work this modification because MinecraftServer doesn't include Queue with futuretask. It fixes if you use viaversion.

# Additional Comments

Just the better performance!

# Checklist:

- [ Y] I have reviewed my code thoroughly.
- [ Y] I have tested my code.
- [ X] My changes generate no new warnings


wuangg:
Have you done any benchmark before saying this PR could give a better performance?

Dysaido:
> Have you done any benchmark before saying this PR could give a better performance?



> I have got an server with 30 players. It was running 3 month. Most process faster with this modification then default calculator, because we don't send the thread to sleep. The yield is a better solution then sleep. It is no coincidence that Mojang switched to this. NetworkManager and ServerConnectin supplemented with isPending and I removed ReentrantReadWriteLock from NetworkManager because it doesn't neccessary. I updated scheduler from new Paper. I removed database from Plugin, because nothing uses this. The benefits of this are clear.

Dysaido:
> Have you done any benchmark before saying this PR could give a better performance?



> I have got an server with 30 players. It was running 3 month. Most process faster with this modification then default calculator, because we don't send the thread to sleep. The yield is a better solution then sleep. It is no coincidence that Mojang switched to this. NetworkManager and ServerConnectin supplemented with isPending and I removed ReentrantReadWriteLock from NetworkManager because it doesn't neccessary. I updated scheduler from new Paper. I removed database from Plugin, because nothing uses this. The benefits of this are clear.

LewUwU:
Can't you do another PR to do only the ticker thing? instead of combining every change (Replace to lambda, channel listener, adventureapi and etc)

Dysaido:
> Can't you do another PR to do only the ticker thing? instead of combining every change (Replace to lambda, channel listener, adventureapi and etc)

I need an guidance. Because I don't have enough experience about git. 😁

sadcenter:
Simply, divide this PR into parts. So, don't create PR with things unrelated to each other
For example:
Instead of creating: `New ticker and NetworkManager`
Create: `New Ticker`, `Updated NetworkManager`
You can do it by using different branches.

PR #384, opened 2022-03-24
fast Versioning

crafter23456:
# Description

Unnecessary to fetch it every startup again, because we stay with 1.8.8 RC2-SNAPSHOT.

# Checklist:

- [x] I have tested my code.

### DIFF
--- a/NachoSpigot-Server/src/main/java/org/bukkit/craftbukkit/util/Versioning.java
+++ b/NachoSpigot-Server/src/main/java/org/bukkit/craftbukkit/util/Versioning.java
@@ -2,6 +2,6 @@ package org.bukkit.craftbukkit.util;
 
 public final class Versioning {
     public static String getBukkitVersion() {
-        return "1.8.8-R0.2";
+        return "1.8.8-R0.2-SNAPSHOT";
     }
 }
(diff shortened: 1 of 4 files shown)


NachoSpigot-Server/src/main/java/org/bukkit/craftbukkit/util/Versioning.java
```diff
@@ -1,29 +1,7 @@
...
-                properties.load(stream);
-
-                result = properties.getProperty("version");
-            } catch (IOException ex) {
-                Logger.getLogger(Versioning.class.getName()).log(Level.SEVERE, "Could not get Bukkit version!", ex);
-            }
-        }
-
-        return result;
+        return "1.8.8-R0.2-SNAPSHOT";
```
Elierrr, line 5: @Lucaskyy Should we finally switch to `1.8.8-R0.1-SNAPSHOT` like upstream?

Elierrr (approved):

NachoSpigot-Server/src/main/java/org/bukkit/craftbukkit/util/Versioning.java
LewUwU: R0.1 is the most used

Lucaskyy: @crafter23456 ^

Elierrr (changes requested):
Update the version in the pom.xml as well

Elierrr (approved):

Lucaskyy (approved):

PR #385, opened 2022-03-24
Disable Nether Roof Interaction

crafter23456:
I cant break the roof in creative or like that, seems robust, but i can place but not break on top of it and the message appears, yeah because its coded like it should do that. cause of my other PRs, the readme order will change anyways so i pasted it just on the bottom of the flamepaper patches

# Checklist:

- [x] I have tested my code.

### DIFF
--- a/NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
+++ b/NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
@@ -143,7 +143,7 @@ public class NachoConfig {
         c.addComment("settings.commands.enable-help-command", "Toggles the /help command");
         c.addComment("settings.use-improved-hitreg", "Enables the usage of an improved hitreg based on lag compensation and small other details.");
         c.addComment("settings.disable-disconnect-spam", "Disables that players can be kicked because of disconnect.spam.");
-        c.addComment("settings.disable-nether-roof-destroying", "Disables destorying the nether roof.");
+        c.addComment("settings.enable-nether-roof-destroying", "Enables a bigger protection for the nether roof.");
         NachoWorldConfig.loadComments();
     }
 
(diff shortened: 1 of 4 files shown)


NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
```diff
@@ -404,4 +405,10 @@ private static void enableImprovedHitReg() {
     private static void disableDisconnectSpam() {
         disableDisconnectSpam = getBoolean("settings.disable-disconnect-spam", false);
     }
+
+    public static boolean disableNetherroofDestroying;
```
Elierrr, line 409: Incorrect name, the config disabled placing on nether roof not destroying it

NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
```diff
@@ -143,6 +143,7 @@ static void loadComments() {
         c.addComment("settings.commands.enable-help-command", "Toggles the /help command");
         c.addComment("settings.use-improved-hitreg", "Enables the usage of an improved hitreg based on lag compensation and small other details.");
         c.addComment("settings.disable-disconnect-spam", "Disables that players can be kicked because of disconnect.spam.");
+        c.addComment("settings.disable-nether-roof-destroying", "Disables destorying the nether roof.");
```
Elierrr, line 146: Update comment

Elierrr (changes requested):

Elierrr:
I don’t really see the point of this since it can very easily be done with a plugin

NachoSpigot-Server/src/main/java/me/elier/nachospigot/config/NachoConfig.java
```diff
@@ -404,4 +405,10 @@ private static void enableImprovedHitReg() {
     private static void disableDisconnectSpam() {
         disableDisconnectSpam = getBoolean("settings.disable-disconnect-spam", false);
     }
+
+    public static boolean enableNetherroofDestroying;
```
Elierrr, line 409: It should be `disableNetherroofInteraction`

Elierrr (changes requested):

Elierrr (approved):
This PR can be merged, however I don’t believe this is necessary since it can easily be done with a plugin

Lucaskyy:
I agree with Elier, plugins should handle this. NachoSpigot is for performance, not features plugins already do. Thanks anyway though!

PR #389, opened 2022-03-27
Backport Material and Locale APIs

Elierrr:
# Description

Backports some Bukkit Material and Locale APIs

# Checklist:

- [x] I have tested my code.

### DIFF
--- a/NachoSpigot-API/src/main/java/org/bukkit/event/player/PlayerLocaleChangeEvent.java
+++ b/NachoSpigot-API/src/main/java/org/bukkit/event/player/PlayerLocaleChangeEvent.java
@@ -4,35 +4,26 @@ import org.bukkit.entity.Player;
 import org.bukkit.event.HandlerList;
 
 /**
- * Called when the locale of the player is changed.
+ * Called when a player changes their locale in the client settings.
  */
 public class PlayerLocaleChangeEvent extends PlayerEvent {
+
     private static final HandlerList handlers = new HandlerList();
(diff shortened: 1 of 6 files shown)


LewUwU (approved):
LGTM

NachoSpigot-API/src/main/java/org/bukkit/event/player/PlayerLocaleChangeEvent.java
```diff
@@ -4,35 +4,26 @@
...
-        this.newLocale = newLocale;
-    }
+    private final String locale;
 
-    /**
-     * Gets the locale the player switched from.
-     *
-     * @return player's old locale
-     */
-    public String getOldLocale() {
```
Lucaskyy, line 25: One more thing, please add these methods back and just return the `locale` variable for backwards compatibility.

Lucaskyy (changes requested):

NachoSpigot-API/src/main/java/org/bukkit/event/player/PlayerLocaleChangeEvent.java
Elierrr: I don’t think it’s necessary. This event was implemented during a time when not a lot of people used Paper and plugin developers wouldn’t use it’s API and in 1.9, the package for this event would be moved to Paper’s new package.

Lucaskyy (approved):

Issue #176, opened 2021-10-09
SERIOUS ERROR WITH MYSQL


samupro-dev:
After you updated the jar my server no longer connects to mysql, I tried another fork and it works. Please settle down as soon as possible!

samupro-dev:
gives me these errors in console:
```
java.lang.RuntimeException: Failed to get driver instance for jdbcUrl=jdbc:mysql://127.0.0.1:3306/database1
	at com.andrei1058.bedwars.libs.hikari.util.DriverDataSource.<init>(DriverDataSource.java:114) ~[?:?]
	at com.andrei1058.bedwars.libs.hikari.pool.PoolBase.initializeDataSource(PoolBase.java:325) ~[?:?]
	at com.andrei1058.bedwars.libs.hikari.pool.PoolBase.<init>(PoolBase.java:114) ~[?:?]
	at com.andrei1058.bedwars.libs.hikari.pool.HikariPool.<init>(HikariPool.java:108) ~[?:?]
	at com.andrei1058.bedwars.libs.hikari.HikariDataSource.<init>(HikariDataSource.java:81) ~[?:?]
	at com.andrei1058.bedwars.database.MySQL.connect(MySQL.java:87) ~[?:?]
	at com.andrei1058.bedwars.BedWars.onEnable(BedWars.java:353) ~[?:?]
	at org.bukkit.plugin.java.JavaPlugin.setEnabled(JavaPlugin.java:321) ~[server.jar:git-NachoSpigot-"4839ac6"]
	at org.bukkit.plugin.java.JavaPluginLoader.enablePlugin(JavaPluginLoader.java:332) [server.jar:git-NachoSpigot-"4839ac6"]
	at org.bukkit.plugin.SimplePluginManager.enablePlugin(SimplePluginManager.java:393) [server.jar:git-NachoSpigot-"4839ac6"]
	at org.bukkit.craftbukkit.v1_8_R3.CraftServer.loadPlugin(CraftServer.java:374) [server.jar:git-NachoSpigot-"4839ac6"]
	at org.bukkit.craftbukkit.v1_8_R3.CraftServer.enablePlugins(CraftServer.java:333) [server.jar:git-NachoSpigot-"4839ac6"]
	at net.minecraft.server.v1_8_R3.DedicatedServer.init(DedicatedServer.java:205) [server.jar:git-NachoSpigot-"4839ac6"]
[...]

PR #396, opened 2022-04-03
Cleanup

128931:
# Description

Cleaned up a decent amount of the API code to be up to Java 8 standards.

# How has this been tested?

I have built the jar and ran it on localhost I saw no new errors/exceptions I joined and played mc for approx 10 mins doing various things an mc player would do and I saw nothing out of the ordinary happen

# Additional Comments

This is my first time working on a spigot/fork of it so if I have changed something that would break a feature or something within spigot lmk

# Checklist:

- [x] I have tested my code.


HeathLoganCampbell (commented):
Do you have benchmarkes to back this up? because that doesn't make sense to set them to zero

128931:
> Do you have benchmarkes to back this up? because that doesn't make sense to set them to zero, because that is a pretty big claim

no not currently I could revert it if wanted

NachoSpigot-API/pom.xml
```diff
@@ -65,12 +65,14 @@
         <dependency>
             <groupId>org.yaml</groupId>
             <artifactId>snakeyaml</artifactId>
-            <version>1.15</version>
+            <!-- 1.25 and below allows for https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-18640 -->
+            <version>1.30</version>
```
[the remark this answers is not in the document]
128931: I could try out a lot of popular plugins on my localhost server and see if they break or not if u want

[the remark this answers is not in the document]
128931: > All of my plugins didnt break. Thats not the point, but some very old and badly coded plugins will stop working. thats the point

I mean ngl ppl probably shouldn't be using outdated and badly coded plugins but I already know ppl are gonna be doing that anyway so I can convert it back to 1.15 if wanted

NachoSpigot-Server/pom.xml
```diff
@@ -80,59 +80,60 @@
...
+            <version>3.36.0.3</version>
             <type>jar</type>
             <scope>runtime</scope>
         </dependency>
         <dependency>
             <groupId>mysql</groupId>
             <artifactId>mysql-connector-java</artifactId>
-            <version>5.1.14</version>
+            <!-- < 8.0.16 versions are affected by https://cwe.mitre.org/data/definitions/843.html -->
+            <version>8.0.28</version>
```
crafter23456, line 105: A bump could cause very old and badly coded plugins to stop working. so i wouldnt go over 5.1.49 but dunno if that still breaks things

NachoSpigot-Server/pom.xml
```diff
@@ -80,59 +80,60 @@
...
-            <version>2.12.1</version>
+            <version>2.14.6</version>
             <type>jar</type>
             <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>org.xerial</groupId>
             <artifactId>sqlite-jdbc</artifactId>
-            <version>3.7.2</version>
+            <version>3.36.0.3</version>
```
crafter23456, line 97: Same like with the mysql connector

128931: I get the mysql connector part but this one is kinda ridiculous because version 3.7.2 is literally from 2010

NachoSpigot-API/pom.xml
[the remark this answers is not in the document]
crafter23456: guess a spigot fork made that configurable but that would explode the jar size😄i would revert it even if it would be better for 90% of the user. if elier doesnt have any other solution to solve it

CyberFlameGO:
I recall trying to update some of these myself; it broke many things

NachoSpigot-Server/pom.xml
crafter23456: we know :/ updating yaml, mysql and sqlite stuff causes always issues of that badly old coded plugins

LewUwU: > I get the mysql connector part but this one is kinda ridiculous because version 3.7.2 is literally from 2010

When you create plugins based on a driver of 2010, that's why you can't just bump it without testing with some common plugins

Elierrr: #176

128931:
It's 6 am and I'm really tired so I'm going to bed my Discord is onetwoeight#0601 in case u need to harass me about more stuff I did

NachoSpigot-API/pom.xml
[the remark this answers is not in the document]
Lucaskyy: Well, one thing we shouldn't forget. This breaks old plugins, right? Yeah, how old is 1.8 again? A lot of 1.8 plugins are no longer being updated, which means they'll break if we update the dependencies they use to newer versions.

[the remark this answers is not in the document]
crafter23456: > Well, one thing we shouldn't forget. This breaks old plugins, right? Yeah, how old is 1.8 again? A lot of 1.8 plugins are no longer being updated, which means they'll break if we update the dependencies they use to newer versions.

Most plugins support 1.8-1.18. The plugins I need arent only 1.8, but yes some old odd plugins can be only for that version.

Ytnoos:
Why did you revert?

128931:
I would clean up the server directory rn but the code gives me nightmares due to it being decompiled code

128931:
I'm gonna ignore cleaning up the NMS for a bit as I've looked thru some of the classes and I almost got traumatized looking at that decompiled code

NachoSpigot-Server/src/main/java/net/techcable/tacospigot/HopperPusher.java
```diff
@@ -1,15 +1,10 @@
 package net.techcable.tacospigot;
 
-import net.minecraft.server.AxisAlignedBB;
-import net.minecraft.server.BlockPosition;
-import net.minecraft.server.Entity;
-import net.minecraft.server.MathHelper;
-import net.minecraft.server.TileEntityHopper;
-import net.minecraft.server.World;
+import net.minecraft.server.*;
```
crafter23456, line 3: is it that good to cover them all with *?

128931: I believe that it's better practice to do .* if u have 5 or more imports of the same directory

crafter23456: normally i would say yes but net.minecraft.server is a bunch of classes. or is that irrelevant for compiling?

128931: What's ur point on "net.minecraft.server is a bunch of classes"

crafter23456: that it takes a bit more time to go through the classes, but i think thats irrelevant, right?

128931: As far as I know it affects nothing performance wise

Elierrr: Java byte code uses full class names not imports, so it wouldn’t affect performance

